PR #1, opened 2023-10-23
height slider

LankLTE:
yeah. Ideally it'd use scale but that was such a nightmare I'll just leave it be for now.


LankLTE:
To anyone merging this downstream, you’ll also want to merge commit d97b20a8a89e68f660f913ce96f91cea2aea6d6b which fixes various issues but wasn’t done through a PR (oops).

PR #8, opened 2023-11-03
painful chemicals

CrigCrag:
Chemical effects pass.

## About the PR

Changed a bunch of chemicals to make them feel interactive. Acids deal much more damage while splashed on people, because they're acids and the old damage values were really underwhelming even when throwing bluespace beakers or using backpacks full of acid to shoot acid at people. Every element that is caustic or irritable in real life is now dangerous to spill or slip in. Lab coats and PPE, people.

Amatoxin was also reworked to be a slower killer. 9u will crit someone but it will take about 100 seconds. Dying silently really fast via poison isn't great for RP in my opinion and Amatoxin as always been overkill. It now serves as a bridge between lead and other poisons.

Bio suits have been buffed to help protect janitors and others from chemical threats.
## Why / Balance

HRP, chemicals are dangerous even if you aren't chugging them. This should add some sense of danger when working with chemicals. Janitors with PPE should be called to clean up acid spills and chemical spills should be avoided. Inert chemicals from the dispenser are now a minority. Chemists should be careful when transporting or handling most chemicals, and lab coats are no longer a complete afterthought.
## Media

## Breaking changes

**Changelog**

:cl:
- tweak: The majority of chemical elements are now dangerous when in contact with skin. Wear a lab coat and be careful of unidentified spills!
- tweak: Amatoxin has been reworked and no longer kills incredibly quickly.
- tweak: Acid is now much more dangerous when spilled on skin.
- tweak: Bio suits are now more effective against chemical spills.

### DIFF
--- a/Resources/Prototypes/Reagents/toxins.yml
+++ b/Resources/Prototypes/Reagents/toxins.yml
@@ -407,12 +407,12 @@
   color: "#D6CE7B"
   metabolisms:
     Poison:
-    metabolismRate: 0.1
-      effects:
-      - !type:HealthChange
-        damage:
-          types:
-            Poison: 1
+    metabolismRate: 0.10
+    effects:
+    - !type:HealthChange
+      damage:
+        types:
+          Poison: 1
 
 - type: reagent
   id: VentCrud
--- a/Resources/Prototypes/Reagents/toxins.yml
+++ b/Resources/Prototypes/Reagents/toxins.yml
@@ -407,12 +407,12 @@
   color: "#D6CE7B"
   metabolisms:
     Poison:
-    metabolismRate: 0.10
-    effects:
-    - !type:HealthChange
-      damage:
-        types:
-          Poison: 1
+      metabolismRate: 0.90
+      effects:
+      - !type:HealthChange
+        damage:
+          types:
+            Poison: 1
 
 - type: reagent
   id: VentCrud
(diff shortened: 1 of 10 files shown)


CrigCrag:
oops

CrigCrag:
test fail is crying about cargo sell prices that's _probably_ fine

LankLTE:
> test fail is crying about cargo sell prices that's _probably_ fine

Fix the test fails since it'll scream constantly if not. Setting the price to 850 should be a fine solution to that.

CrigCrag:
no more test fail

LankLTE:
i'll read over this tomorrow to make sure its' all okay

Resources/Prototypes/Reagents/toxins.yml
```diff
@@ -395,11 +407,12 @@
   color: "#D6CE7B"
   metabolisms:
     Poison:
+      metabolismRate: 0.90
```
LankLTE, line 410: MetabolismRate is in units per second, with 0.5 being the base. This is almost twice as fast as normal which I assume you didn't intend.

LankLTE (dismissed):
Other than amatoxin I think it's fine. Acid buffs might be overdone but they're hardly used anyways so I'm down to try them.

Resources/Prototypes/Reagents/toxins.yml
CrigCrag: oops, meant to put 0.09

PR #16, opened 2023-11-04
Increase Time To Rot (The Rot No Longer Consumes)

### LINKED ISSUE #3: [Feature Request/Idea Discussion] Antithesis Necrosis Casket (The ANC)/Increasing Time To Rot
### Description
There was some discussion in the discord about increasing the rot time due to the length rounds are planned to be. Then the discussion thanks to Nairrod shifted to what a machine that reverses the rot status effect would look like. I decided to move it to a github issues to keep the discussion more organized and to not lose the idea.

Nairrod came up with the name found [here](https://github.com/cosmatic-drift-14/cosmatic-drift/issues/3#issuecomment-1783408087)

**Idea Breakdown**
The intent is a machine that takes a while to revert the decomposition of a body, through use of a chemical (Probably a new one? Epi could work aswell if we don't want to add another chemical for chemists to make).

Insert body into machine -> Insert chemical -> Turn on machine -> "x" amount of time later, the body is no longer decomposed and can be revived normally.

**Balancing**
This should probably take a while, probably five or more minuets, maybe even longer?
This should still cause memory loss for the shift most likely, the IC reasoning could be the process doesn't entirely protect the brain, or the brain damage from rotting is to severe to preserve the recent (shift) memories.
This should probably be a Tier two or three science tech, I'm thinking three because it does out right bring people back into the shift.
The further rotted a body is rotted, it should take damage per unit of rot removed. Probably lacerations and heat. The more rotted the body is, the more damage will have to be healed after the machine is done. [Idea was put here.](https://github.com/cosmatic-drift-14/cosmatic-drift/issues/3#issuecomment-1783408087)

**Implimation/Other Notes**
Besides spriting and the actual code work, we should probably include a guidebook entry for this.

PursuitInAshes:
## About the PR
Oh boy, don't you love untested changes? I love untested changes.
closes #3 
This increases the rot timer to take longer overall, I'm looking for input as to how much longer it should be increased, as I'm going with five extra minuets for now. We could possibly bring it to twenty over all, but I don't know if that is quite justified.
**The rot consumes slower than before.**

## Why / Balance
Due to the increased round duration, I believe the time to rot should reflect that, with standard round times (1H 20M) it takes you about 12% of the round to rot, but with out extended round times (5 hours), that drops to 0.03. It should take more time for people to rot. I doubt there will be many times people actually rot due to antags, but I still think this should be increased not only because of antags, but the chances of accidents that can occur during the shift.

## Technical details
Changes line 16 of PerishableComponent.cs this should hypothetically add five minuets to the rot timer, bringing it to fifteen total.

I will include media upon request (assuming I can actually figure out how to test....)

If for whatever reason this doesn't work, I'll actually take a closer look.

:cl: 
 tweak:  The time for bodies to rot has been increased from 10 -> 15 minuets.

### DIFF
--- a/Content.Shared/Atmos/Miasma/PerishableComponent.cs
+++ b/Content.Shared/Atmos/Miasma/PerishableComponent.cs
@@ -13,7 +13,7 @@ public sealed partial class PerishableComponent : Component
     /// How long it takes after death to start rotting.
     /// </summary>
     [DataField("rotAfter"), ViewVariables(VVAccess.ReadWrite)]
-    public TimeSpan RotAfter = TimeSpan.FromMinutes(10);
+    public TimeSpan RotAfter = TimeSpan.FromMinutes(15);
 
     /// <summary>
     /// How much rotting has occured


LankLTE:
I don't think changelogs actually work, so **denied**

PR #38, opened 2023-11-12
Add additional HoS Winter Coat matching Warden's

PursuitInAshes:
## About the PR
Changes the HoS's winter coat to match the shape of the warden winter coat, but keep the same color palet.

## Why / Balance
A lot of discord discussion can be found [here](https://discord.com/channels/1163352113162768436/1163713936231764059/1173093417467789363). Though the short version is some felt as if the HoS's winter coat should match the wardens. Plus, I needed more practice with spriting so I can hopefully work on some bigger projects.

## Media
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase.

Icon:
Content Client_5GwNjfpKVD

Hoods Down & Dropped On Floor Sprite
Content Client_EqpiGzNyYK

Worn With Other Security Equipment
Content Client_gtRvgwrLoQ

Hoods
Content Client_OWwrTqrQ1a


Gotimanga:
I like the new sprite for the coat, however don't you think you should buff the amour values similar to the warden's coat?
They now look very similar, and it will be quite misleading if a coat that looks armoured retains the original protection from a normal winter coat while the warden's looks very similar, but is substantially more resistant to some damage types.

LankLTE (changes requested):
The old winter coat should stay in some form, probably just available through the uniform printer.

LankLTE (approved):

PR #51, opened 2023-11-22
Tune Events

LankLTE:
Makes some of the more disruptive ones less common, and makes events on average happen less often.

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -15,7 +15,7 @@
   noSpawn: true
   components:
   - type: StationEvent
-    weight: 5
+    weight: 8
     startDelay: 30
     duration: 35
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/GameRules/events.yml
```diff
@@ -175,8 +178,7 @@
       path: /Audio/Announcements/attention.ogg
     startDelay: 10
     earliestStart: 30
-    minimumPlayers: 35
```
Nairodian, line 178: I'd be careful removing this from MouseMigration since it also includes ratking spawns, which can be pretty fucking devastating to smaller crews. VentCritters is _just_ mice. (Which also has a 15 min pop, what?)

LankLTE: Glad someone reads my PRs.

PR #53, opened 2023-11-23
Fix Stacking Bug

LankLTE:
Known issue that the UI doesn't update, no idea why. Will worry about it later.


LankLTE:
> I meant posting an issue for it, but I suppose this works too.

I'm still fixing the first bit. Can post an issue about the second bug if you want still.

PR #57, opened 2023-12-01
Bicaridine and Brute Rework

CrigCrag:
## About the PR

GONE are the days of medical doctors injecting bicaridine... dermaline... all that jazz.. without ANY THOUGHT WHATSOEVER. The only thing they've ever had to think about was overdosing. This PR introduces five new reagents:
1: Razorium: A non-newtonian fluid that becomes very sharp when force is applied. Do your best not to slip in this, as it's like falling in razor blades. This chemical is produced by a reaction between **Bicaridine and Dermaline.** If you wish to heal brute and burn at the same time, you'll have to resort to some of the new, more specific (and less reactive) chemicals.
2: Puncturase: A fizzy chemical that is very effective at healing piercing damage! However, it does leave some blunt tissue damage in its wake.
3: Bruizine: Originally developed as cough medicine, scientists realized that bruizine is very good at treating blunt-force trauma!
4: Lacerinol: A fun little chemical that boosts collagen synthesis to heal slash damage.
5: Agitaseptici: A strange, airborne space bacteria that was recently discovered to be incredibly resilient and incredibly good at contaminating stuff. Luckily, it's almost always dormant, until it detects a sweet mixture of medication to feed off of. **Agitaseptici will be produced when you mix any two brute medications. (ex. Bicaridine and lacerinol, or puncturase and bruizine.)** This bacteria isn't super dangerous, but it should be removed from your bloodstream with charcoal or ipecac as soon as possible to avoid complications. 
## Why / Balance

Medical is kind of boring and chemistry doesn't have much to do. Doctors only really think about overdosing, this should add a bit of strategy into the mix. Doctors must be much more keen about their dosages, lest they risk an Agitaseptici infection. However, if they are successful, they may find their patient recovering faster than if they'd used Bicaridine.
Through this and some other PRs I hope to add some depth to chemistry in a somewhat interesting way using chemicals that probably couldn't be trusted in the hands of your average LRP powergamer. The end goal is to also provide new ways for danger and conflict to arise without any major antags.
## Technical details

I added a bunch of reagents and flavors it's not really too interesting.
## Media

https://github.com/cosmatic-drift-14/cosmatic-drift/assets/137215465/c2291244-6d70-4d8c-af2e-fb5fa960593f

**Changelog**

Make sure to take this Changelog template out of the comment block in order for it to show up.
:cl:
- add: Added three specific chemicals to target each group of brute damage.
- add: Added Razorium, a sharp and dangerous chemical created when Bicaridine and Dermaline are mixed!
- add: Added Agitaseptici, a strange and mildly dangerous bacterium found infecting mixed brute medication.


CrigCrag:
I will see if i can put together a guidebook entry ive never made one of those before. The chemical section does auto-update, though

LankLTE:
I honestly think that the three specific brute chems could be upstreamed, but that’s not the biggest deal ever. I’ll review this more later.

CrigCrag:
i hope i didnt just explode everything

CrigCrag:
we ball

PR #58, opened 2023-12-01
Advanced Brute Medication Guidebook entry

CrigCrag:
## About the PR
Requires #57


CrigCrag:
haha imagine if i pushed the entire branch again instead of just the guidebook that would be crazy

PR #66, opened 2023-12-08
Medical gauze markings

KittenColony:
## About the PR

Adds gauze markings to all limbs, excluding for Diona.

## Why / Balance

It allows players to RP long-term injuries, blah blah expression of character. You can be a burn victim now

## Media
image_2023-12-08_223123870

**Changelog**

:cl:
- add: Added various gauze wraps to most species in the form of markings


KittenColony:
Diona are too weirdly shaped to easily allow for the gauze to work, plus they're plants, they wouldn't need it anyway

PR #76, opened 2023-12-16
Antidepressants and Anxiety RP Meds

CrigCrag:
I decided to atomize RP meds because they ended up being more complicated than I thought.
## About the PR

This PR adds 6 medicines that are primarily RP-based. Their main purpose is to act as meds that characters may need to take to deal with emotional issues. They give RP prompts that you may or may not follow, which get wackier if you overdose. There are a few cool overdose effects ranging from short-term paralysis to psychosis. They have a lot of interactions and are all pretty fleshed out I think. They all are detailed automatically in the Chemicals section of the guidebook, and can be found by searching for "Anxiety" or "Antidepressant." Their guidebook descriptions detail on how they work fairly well. 

These chemicals are **primarily** for RP value and do not yet provide any mechanical advantage whatsoever.
The rate at which you receive prompts isn't really intrusive unless you are overdosing. 

## Antidepressants
From least impactful to most:
image
image
image
image
(Note that bleach is a catalyst in the above reaction and is not consumed)
image
image

## Anxiety Medication
From least impactful to most:
image
image
image
image
(Note that chloral hydrate is a catalyst in the above reaction and is not consumed)
image
image

## Why / Balance

It seems like there is a lot of demand for RP chemicals. With the mental of in-game characters having a much greater impact on HRP I think that RP chemicals are a good way to start. By design, they create small RP prompts casually from time to time while also giving a bit of guidance for RP if you ever overdose. They do not provide any mechanical benefit whatsoever. Some overdoses might serve somewhat well as a poison maybe? Considering it's an HRP server I doubt it will be a problem + there are other chemicals that do the job better anyways.

## Technical details

Utilizes a lot of reagent effects and interactions, some of which have been previously underutilized.

**Changelog**

:cl:
- add: Added three types of antidepressants and three types of anxiety medication for RP purposes. Psychologist gameplay is real.

### DIFF
--- a/Resources/Locale/en-US/_CD/reagents/RPmeds.ftl
+++ b/Resources/Locale/en-US/_CD/reagents/RPmeds.ftl
@@ -1,0 +1,60 @@
+### Messages to be used with RP-based medication.
+
+reagent-effect-antidepressant-mild1 = You feel like things aren't so bad. 
+reagent-effect-antidepressant-mild2 = The world seems a bit brighter. 
+reagent-effect-antidepressant-mild3 = You feel a bit better about yourself.
+reagent-effect-antidepressant-mild4 = You feel slightly motivated.
+
+reagent-effect-antidepressant-normal1 = You feel pretty alright. 
+reagent-effect-antidepressant-normal2 = You find it hard to focus. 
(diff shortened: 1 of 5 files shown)


KittenColony:
:godo:

PR #103, opened 2024-01-06
remove exterminator

LankLTE:
look at me good enough to actually try tests


CrigCrag:
rename pr to exterminates exterminator

LankLTE:
going to do this properly later

PR #105, opened 2024-01-07
Anti-Psych RP Chem

dffdff2423:
## About the PR

 We have been using various mixes of chemicals for an RP anti-psychotic for quite some time. This turns the most common mix we use into an actual chemical.  It also suppresses paracusia in almost the same way Diphenylmethylamine suppresses narcolepsy as I figured it was appropriate.

## Why / Balance

Should not really affect game balance in any way as it is an RP chem. Technically it is possible to put someone to sleep with it's overdose but chloral hydrate is a far more effective solution.

The name Saphrodipine was chosen mostly at random. I wanted to avoid using the name Haloperidol (what we have been using for this mixture IC) because it is apparently a real medication and I figure it is best to avoid such things.

## Technical details
The chem functions basically the same the other RP chems. To be able to suppress paracusia I added the ResetParacusia reagent effect which required modifications to ParacusiaComponent to sync the delay to the client because paracusia is mostly handled client side.  

## Media
guidebook

**Changelog** 
(please credit to aquif)

:cl:
- add: Saphrodipine an anti-psychotic RP chem. Also suppresses paracusia.

### DIFF
--- a/Content.Client/Traits/ParacusiaSystem.cs
+++ b/Content.Client/Traits/ParacusiaSystem.cs
@@ -58,6 +58,9 @@ public sealed class ParacusiaSystem : SharedParacusiaSystem
         var timeInterval = _random.NextFloat(paracusia.MinTimeBetweenIncidents, paracusia.MaxTimeBetweenIncidents);
         paracusia.NextIncidentTime += TimeSpan.FromSeconds(timeInterval);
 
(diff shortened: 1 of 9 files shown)


LankLTE:
Both of these look good to me, but I'll review them more thoroughly tomorrow just to be sure.

PR #106, opened 2024-01-07
Extend Round End Time

Nairodian:
## About the PR
Changes the default round end time from 2 minutes to 5 to give more time to tie up RP ends.

## Why / Balance
Feels like a good number, enough time to settle down but not too long that it's obnoxious. I've done this manually several times here and on Upstream, people seem to like it.

## Technical details
Changes the default of the Cvar instead of using the Cvar because the Cvar was last known to crash servers at round end.

**Changelog**
The delay from when the round ends to when the game is returned to the lobby has been upped from 2 to 5 minutes.


LankLTE:
Remind me to revert this when it breaks everything

PR #112, opened 2024-01-11
Larger Character Descriptions

DEATHB4DEFEAT:
# About the PR

Changed the limit on character descriptions from 512 to 1024 characters.

# Why / Balance

512 characters is not a very large amount, especially for an HRP server.

# Changelog

:cl:
- tweak: Character descriptions can now be up to 1024 chars in length

### DIFF
--- a/Content.Shared/Preferences/HumanoidCharacterProfile.cs
+++ b/Content.Shared/Preferences/HumanoidCharacterProfile.cs
@@ -25,7 +25,7 @@ namespace Content.Shared.Preferences
     public sealed partial class HumanoidCharacterProfile : ICharacterProfile
     {
         public const int MaxNameLength = 32;
-        public const int MaxDescLength = 2048; // CosmaticDrift-LargerCharacterDescriptions // Was 512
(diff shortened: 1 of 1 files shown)


dffdff2423:
Honestly I am not sure that larger character descriptions are a good idea. It is already quite hard to read descriptions at the current max length as the text box only stays up when you are in close proximity to the character.

evilexecutive:
> Honestly I am not sure that larger character descriptions are a good idea. It is already quite hard to read descriptions at the current max length as the text box only stays up when you are in close proximity to the character.

This definitely isn't my experience. The text box could be at least three times as long as it is right now, and I'd still be able to comfortably read it. Right now with the painfully short length, I can really only manage a single good paragraph, when I'd rather have at least two. I had just rewritten Farida's description in a doc file to better suit her, and getting just what I consider the basics(by my standards) took 1080 characters before any edits.

LankLTE (changes requested):
This should be set to double (1024) or have the character description you showed off earlier ported. I agree with aquif that, with the current UI, having long character descriptions (especially 2048 characters) is going to be difficult to read.

LankLTE (approved):

PR #113, opened 2024-01-11
Remove Senior Officers Roundstart Secglasses

PursuitInAshes:
## About the PR
The senior officer role now spawns with sunglasses instead of security sunglasses.

## Why / Balance
Secglasses are meant to be a T2 research and have been removed from the roundstart kits of the rest of security. SO's got overlooked due to the nature of how we have been reverting their removal.

**Changelog**
tweak: Senior Officers now spawn with sunglasses instead of security sunglasses, this brings them in line with the rest of roundstart security.


LankLTE:
Not including the CL since it's a relatively minor thing. Hope that's alright.

PR #115, opened 2024-01-11
Disable Clerical and Terminator Events

Nairodian:
## About the PR
Disables the terminator and clerical error events. (And probably doesn't fail tests.)

## Why / Balance
Terminator is a obvious since it's whole goal is to perma kill someone.
Clerical Error could be a up for a bit more debate, given it's nothing major but if I recall correctly we were leaning towards wanting to disable this one.

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -55,17 +55,17 @@
     # duration: 1
   # - type: BureaucraticErrorRule
 
-- type: entity
-  id: ClericalError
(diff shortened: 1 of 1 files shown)


LankLTE:
The main argument for the clerical error being added upstream was so that "someone not being on the manifest doesn't instantly make them valid," but I kinda feel like we can trust our players to not instantly arrest & perma someone for not being on the manifest. The event isn't necessary, basically, and only serves to make gameplay more difficult / obscured.

PR #120, opened 2024-01-13
Ferrous Update

PursuitInAshes:
## About the PR
I don't know why the diff is so big, here is the changelog:

AME Core is now in line with upstream's mapping standards.
More Generators have been added around the map for roundstart power.
The SWAT crate in security has been replaced with one that doesn't start with shotguns.
Disposals now will drop its contents correctly onto the belt instead of into the far wall.
Department Head Cosmetic Dressers have been mapped.
The outer armory windoor is now security access instead of armory access.
The inner bridge airlock is now glass.
Engineering has an engidrobe (Finally).
New Warp Points/Map Beacons have been added, I tried to convert over as many as I could, but I might have missed some.
A clown/mime room has been added above theater, its not a large space, but they can now fill out important documents.

## Why / Balance
Map Maintainer Moment.

Changelog:
Ferrous has received another pass over, all feedback from 1/12/23 and before has been adressed.

### DIFF
--- a/Resources/Maps/_CD/ferrous.yml
+++ b/Resources/Maps/_CD/ferrous.yml
@@ -11850,13 +11850,9 @@ entities:
   entities:
   - uid: 12866
     components:
-    - flags: InContainer
-      type: MetaData
-    - parent: 12865
+    - pos: 20.451284,14.26887
+      parent: 2
       type: Transform
-    - canCollide: False
-      type: Physics
-    - type: InsideEntityStorage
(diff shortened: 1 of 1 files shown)


LankLTE:
This may conflict with the upstream merge since I had to modify the mapping file there, just warning you now. It's a one line change so shouldn't be anything too significant.

LankLTE:
today I will merge map PR without opening it :clueless:

PR #126, opened 2024-01-18
Make Quick Threads Restock-able

Nairodian:
## About the PR
Adds the Quick Threads to the wardrobe restock crate.

## Why / Balance
Should have some way to restock it. (Mostly because I like to keep it stocked as Central Command)


Nairodian:
I uhh, believe the test fail may be unrelated.

Issue #122, opened 2024-01-15
Additional Prisoner Join Message


PursuitInAshes:
## Description
When a prisoner joins (or late joins) an additional message should be sent into the security radio, much like when science researches something. This will serve as an additional "net" to make sure security don't forget about their prisoners once the shift has already started.

## Reasoning
While a message is already sent notifying the station that someone joined the round, it can be very easy to miss during a busy moment in the shift. A red text, which players are more likely to look at/treat as a priority to read (especially as sec), will make sure that when someone is arguing over common that prisoners don't get lost in the chat.

PR #132, opened 2024-01-22
Add prisoner join notification for security radio

dffdff2423:
## About the PR
Implements https://github.com/cosmatic-drift-14/cosmatic-drift/issues/122

One possible issue is it looks like the prisoner is speaking over security radio, but this is applied before the shift assignment message so I doubt it would case too much confusion. I did this because the only entity that is readily available for JobSpecials is the character that is joining.

## Technical details

Add a new JobSpecial called NotifyDepartmentSpecial that notifies a radio channel with the given text upon the job joining. This is then added to the prisoner role.

## Media

prisoner-join

## Breaking changes

**Changelog**

:cl:
- add (aquif): A notification is now played over security radio when a prisoner is assigned to the shift.

### DIFF
--- a/Content.Server/_CD/Roles/NotifyDepartmentSpecial.cs
+++ b/Content.Server/_CD/Roles/NotifyDepartmentSpecial.cs
@@ -1,4 +1,5 @@
 using Content.Server.Radio.EntitySystems;
+using Content.Server.Station.Systems;
 using Content.Shared.Radio;
 using Content.Shared.Roles;
 using Robust.Shared.Prototypes;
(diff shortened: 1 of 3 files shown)


Nairodian:
I'm not too familiar with the code but is there any way to change the name of what's saying it to something static, or perhaps like the name of a grid (terminal grid in this case)?

dffdff2423:
When I originally made the PR I did not know how to do that, but I figured it out while working on records today. 
image

I was not entirely sure how to enable the arrivals terminal so it might display the name of the arrivals terminal with that enabled. Ether way, this is probably better.

dffdff2423:
What I ended up doing was just add one notification per station. I don't exactly know if this would display more than one notification if there is more than one station, or only show the notification for each station on the given station. Ether way, I don't think it matters as no maps have more than one station as far as I know.

Nairodian (approved):
From a visual standpoint I like it. I can't really comment codewise.

PR #141, opened 2024-02-04
Cowboy/Brown Duster

PursuitInAshes:
## About the PR
Adds a new clothing item to help players complete the "Cowboy" aesthetic. I'm iffy about how they are currently accessible, but I'm not sure if there is a better place to add them. They are also accessible from the clothing loose loot spawner.

The metafile is as detailed as I can make it, TG's file organization is sometimes not incredible, covered bases as best I could.

## Media
Content Client_W7BTbgfRf1

**Changelog**
add: A brown duster has been added, it can be found in the quickthreads.


Resources/Prototypes/_CD/Entities/Clothing/OuterClothing/coats.yml
```diff
@@ -0,0 +1,11 @@
...
+- type: entity
+  parent: ClothingOuterStorageBase
+  id: ClothingOuterCoatCowboyDuster
+  name: brown duster
+  description: A brown duster, covered in sand.
+  components:
+  - type: Sprite
+    sprite: _CD\Clothing\OuterClothing\Coats\cowboyduster.rsi
+  - type: Clothing
+    sprite: _CD\Clothing\OuterClothing\Coats\cowboyduster.rsi
```
Nairodian, line 11: Test fail is due to you using the wrong type of slash / not \

Nairodian (changes requested):

Nairodian (approved):

PR #143, opened 2024-02-04
Ferrous Update V1.05

Nairodian:
## About the PR
Putting the map through my own set of changes and oversight.

- Dorms and cryo.
- Touched up every part of maints.
- A few more maints rooms.
- Entirely remapped every light to make things more dim and not flooded by emergency lights.
- Screens.
- Exterior makeover.
- Nearly full merc set (non-armor) scattered in maints.
- Lower chemist slots by one.
- I did so many small things I don't even know anymore.

## Media
Ferrous-0


Nairodian:
Think that'll do fine since it's nothing major.

PR #154, opened 2024-02-11
Tabletop Computers

PursuitInAshes:
## About the PR
Adds most of the computers/consoles as a tabletop variant. If anyone wants to add more its quite easy to add them.

## Why / Balance
More options for mappers to use, helps with head offices, bridge, etc.

## Technical details
YML, all tabletop computers have the [tabletop] suffix. They take all components from their respective computers, so any updates to the base computers will be applied to these aswell.

These can be deconstructed for their computer board just like everything else, the only real "issue" is they default to the base computer model, which may cause the occasional visual issue.

These don't collide without a table, I'm not quite certain how to make them properly collide.

## Media
Content Client_EXHxC1omYn
Content Client_R0RrtGGAv8

Content Client_B1LwyMCQ9H
Content Client_LhxsCsVIls
Content Client_B4JiBQm00k
Content Client_xPGPGYeBjr

**Changelog**
Tabletop computers have been added, currently they are only accessible by means of admins. You can expect to see these around as mappers decide to place them.


dffdff2423:
I don't really like how it seems to clip into the table/fall off the edge here:
image

Nairodian (approved):
They change to full computer frames upon being broken so the broken RSI is unused.
They can't be made which is disappointing.
But neither of these are really required, should be done at some point though (I'll toss an issue up at some point if this gets merged). Other than that, looks good and fine by me.

Nairodian:
> I'm unsure if it will be as easy to make them constructible, but I'll also look into it.
I was thinking you could maybe make it so if you crowbar a normal computer frame it turns it into a tabletop version and vice versa, from there it's like the original construction graph but with the table top version as the start.

PR #197, opened 2024-03-16
Non-English Rules

Nairodian:
## About the PR
Adds some clarification to the Non-English rules.

## Why / Balance
Admin consensus.

## Media
image
image


dffdff2423:
This is entirely a contrived question (so feel free to ignore), but how would something like a "ß" (German Eszett) be handled under the new 3.4?

Nairodian:
> This is entirely a contrived question (so feel free to ignore), but how would something like a "ß" (German Eszett) be handled under the new 3.4?

It's English readable enough to be fine, I'd say.

PR #199, opened 2024-03-18
Control Rules

Nairodian:
## About the PR
Sub-bullet of 4.6.
"You cannot take direct, unchallengeable orders from people other than your job superiors."

## Why / Balance
Admin consensus.


Nairodian:
Yes, this also only applies to command and security roles.

PR #205, opened 2024-03-24
Fit the Records Add and Down Buttons to their Container

DEATHB4DEFEAT:
# Description

Joined earlier and saw the corners not filled and it was bothering me

# Media

### Before

firefox_aVYZ3S3YBg

### After

Content Client_UNjobGBfq3


dffdff2423 (approved):
Thanks, must have overlooked that.

PR #208, opened 2024-03-27
no more rubber!

Aexxie:
aspid and kettle (kettle!) no longer have rubbers


Nairodian:
Good April Fools joke. :godo:

PR #209, opened 2024-03-27
woo actually fixed yae (no rubbers on kettle and ferrous fr this time lads)

Aexxie:
w

### DIFF
--- a/Resources/Maps/gemini.yml
+++ b/Resources/Maps/gemini.yml
@@ -124334,20 +124334,6 @@ entities:
     - type: Transform
       pos: 61.8193,-22.449226
       parent: 2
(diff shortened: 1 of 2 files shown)


Resources/Maps/gemini.yml
```diff
@@ -172929,15 +172915,6 @@ entities:
     - type: Transform
       pos: -31.413286,49.43612
       parent: 2
-- proto: WeaponShotgunDoubleBarreledRubber
```
Nairodian, line 172932: I think you're deleting something entirely different here.

Nairodian (changes requested):

Aexxie:
I didnt even see it had a request. Fixed I think!

PR #214, opened 2024-04-06
Vulpkanin Species

PursuitInAshes:
## About the PR
Adds the Vulpkanin species. They deal 3 slash, 2 damage. With a temperature threshold of 320 - 230. They can bark, growl, and howl.

## Why / Balance
More player species is a good thing. It provides freedom and character expression for players. I believe this firmly falls into the "Distinctly not human category" which I believe we want.

## Media
Content Client_UO2FuciaIr

(Nothing of the character creator or sounds yet. Soon TM)

## Notes
I could make the pie sprite properly work but that would add more lines than I'm comfortable with. Readability is important. If a maintainer wants it let me know. I'm pretty indifferent eitherway.

**Changelog**
The Vulpkanin species has been ported over from DeltaV as of their latest openly useable versions (02/17/24), along with some CD specific fixes, updates, and improvements.


LankLTE:
Will review this before the Saturday playtest.

Resources/Audio/_CD/Voice/Talk/license.txt
LankLTE, line 1: Is there a reason for using this instead of an attributions file?

Resources/Audio/_CD/Voice/Vulpkanin/license.txt
LankLTE, line 1: Same as above (but there already is one?)

Resources/Textures/_CD/Mobs/Customization/Vulpkanin/body_markings.rsi/meta.json
LankLTE, line 1: idk where you copied these from, but should probably point to a specific commit? That goes for every file.

LankLTE (commented):
Minor stuff.

PR #216, opened 2024-04-07
More scarfs

PursuitInAshes:
## About the PR
Adds 12 new scarfs, you can find the list in media. Its from the light orange scarf down. I opted to put them in the winterdrobe since thats where the rest of the scarfs are, but I'm more than willing to move them to quick threads.

## Why / Balance
More clothing for player and character representation

## Media
Content Client_5iaycwUCLb

## Breaking changes
Nothing currently, since they are in the winterdrobe they may cause merge conflicts later but I tried to label them decently well.

**Changelog**
Various scarfs have been added, you can find them in the winterdrobe. These are namely pride related scarfs.


Resources/Prototypes/_CD/Entities/Clothing/Neck/scarfs.yml
```diff
@@ -0,0 +1,135 @@
...
+  components:
+  - type: Sprite
+    sprite: _CD/Clothing/Neck/Scarfs/Pride_Scarfs/pan.rsi
+  - type: Clothing
+    sprite: _CD/Clothing/Neck/Scarfs/Pride_Scarfs/pan.rsi
+
+- type: entity
+  parent: ClothingNeckBase
+  id: ClothingNeckScarfStripedNonBinary
+  name: striped nob-binary scarf
```
Nairodian, line 107: I assume that's supposed to be "non".

LankLTE:
Starting to think at this point we should just add the pridevend tbh

PR #225, opened 2024-04-19
Implement Loadouts

LankLTE:
Loadouts + Basic fixes to make it functional, as well as removing senior role loadouts. Anything you want added should be done in a separate PR after this one.


Content.Shared/Preferences/HumanoidCharacterProfile.cs
```diff
@@ -43,13 +50,12 @@ public sealed partial class HumanoidCharacterProfile : ICharacterProfile
...
             Gender gender,
             HumanoidCharacterAppearance appearance,
-            ClothingPreference clothing,
-            BackpackPreference backpack,
             SpawnPriorityPreference spawnPriority,
             Dictionary<string, JobPriority> jobPriorities,
             PreferenceUnavailableMode preferenceUnavailable,
             List<string> antagPreferences,
             List<string> traitPreferences,
+            Dictionary<string, RoleLoadout> loadouts)
```
MciReloaded, line 58: Expects a ; at the end of the line here, might be what's causing issues?

dffdff2423: No, the ) at the end of this line should be a ,

PR #227, opened 2024-04-20
Increase comms console character limit and lower cooldown

dffdff2423:
## About the PR

Title.

## Why
I have seen a ton of people get their message massively cut off. It also often happens
that multiple heads want to make their own announcement at the same time. I
don't think we are that worried about people spamming the console on CD. I also
made an [upstream PR](https://github.com/space-wizards/space-station-14/pull/27145) to properly fix the message being cut off.

## Media

**Changelog**

:cl: Aquif
- tweak: Increase communication console character limit and lower the cooldown.


dffdff2423:
Looks like my editor also automatically trimmed some line ending whitespace. If this is an issue (for merging upstream) I can remove that change.

PR #235, opened 2024-04-26
CD Loadouts Patch

Nairodian:
## About the PR
Patches all the senior, PI, and prisoner roles to have loadouts. (Primarily just their jumpsuit choice due to that being rolled into loadouts.)

Normal medical doctors don't get access to the physician beret anymore.
I also gave PI a grab bag of items that seemed to fit their theme in their loadout, but apart from that touched basically nothing else.

## Why / Balance
Mostly just because not being able to pick between bags or jumpsuit/skirt choice sucks.

**Changelog**
CD roles (seniors, private investigator, and prisoner) now have loadouts.


dffdff2423 (approved):
Please ping me when this gets merged so I can fix my PR.

PR #237, opened 2024-04-27
Fix ETU Loadout

dffdff2423:
## About the PR

- **Mark ETU Pilot Helmet as DEBUG**
    - KittenColony said this was intended to me a debug item
- **Remove ETU pilot helmet from loadouts**
- **Disable ETU harness from loadouts**
    - Someone needs to fill it with the EMT belt gear before it can be used in
      loadouts. In addition, I incorrectly assumed it was a harness when it was
      a belt.


KittenColony:
👍

PR #241, opened 2024-04-29
Stowaway trait

Just-a-Unity-Dev:
## About the PR

Adds the "stowaway" trait, which removes anything within a players PDA slot, alongside spawning them inside of a closet/crate.

## Why / Balance

I thought it would be cool.

## Technical/breaking details

Makes the PlayerSpawningEvent of TraitSystem public, and instead spawns the traits **on** player spawn instead of after (after as in once a spawn has been picked).

## Media

https://github.com/cosmatic-drift-14/cosmatic-drift/assets/67359748/b3f74e92-9ea0-4fc9-9d02-cca85ef8f7ae

Please do your best to ignore the music in the video.

**Changelog**

:cl:
- add: The Stowaway trait has been added. It will spawn you in a random unmarked crate/closet without your PDA.

### DIFF
--- a/Content.Server/Traits/TraitSystem.cs
+++ b/Content.Server/Traits/TraitSystem.cs
@@ -1,6 +1,7 @@
 using Content.Server.GameTicking;
 using Content.Shared.Hands.Components;
 using Content.Shared.Hands.EntitySystems;
+using Content.Shared.Preferences;
 using Content.Shared.Traits;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Serialization.Manager;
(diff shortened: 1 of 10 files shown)


Content.Server/Traits/TraitSystem.cs
```diff
@@ -1,6 +1,7 @@
 using Content.Server.GameTicking;
 using Content.Shared.Hands.Components;
 using Content.Shared.Hands.EntitySystems;
+using Content.Shared.Preferences;
```
dffdff2423, line 4: CD imports should be separated from wizden imports.

Content.Server/Traits/TraitSystem.cs
```diff
@@ -20,45 +21,56 @@ public override void Initialize()
         SubscribeLocalEvent<PlayerSpawnCompleteEvent>(OnPlayerSpawnComplete);
     }
 
-    // When the player is spawned in, add all trait components selected during character creation
-    private void OnPlayerSpawnComplete(PlayerSpawnCompleteEvent args)
+    /// <summary>
+    /// Attempts to add traits to the entity.
+    /// </summary>
+    /// <returns></returns>
+    public bool TryAddTraits(EntityUid mob, HumanoidCharacterProfile characterProfile)
```
dffdff2423, line 28: Is there a less invasive way of doing this? This seems like it would make merges a nightmare.

dffdff2423 (changes requested):
I just did a quick once over and I am just commenting on the code. Other than the things noted above, you should add comments whenever you modify wizden code stating the change is a CD change (eg. `// CD: Stowaway Trait` or `// CD: Entity   Spawn Whitelist`).

dffdff2423:
It would also probably be good to not include them on the manifest or load their records.

Just-a-Unity-Dev:
> It would also probably be good to not include them on the manifest or load their records.

Planned

Content.Server/Traits/TraitSystem.cs
Just-a-Unity-Dev: Not that I can think of

dffdff2423: When I wrote that I did not realize that `OnPlayerSpawnComplete` was private. This is not as big of a deal as I thought it was, but wait for Lank to make the final call.

LankLTE:
After discussion on discord it was decided this is unwanted in it's current implementation.

PR #244, opened 2024-05-01
Double SMES capacity 

LankLTE:
Jank mobile webedit 2
Necessary for our arrivals system to prevent power from running out roundstart. Also better to prevent constant blackouts.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Power/smes.yml
+++ b/Resources/Prototypes/Entities/Structures/Power/smes.yml
@@ -32,7 +32,7 @@
     - type: Appearance
     - type: Battery
       startingCharge: 0
-      maxCharge: 8000000
+      maxCharge: 16000000
     - type: ExaminableBattery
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Entities/Structures/Power/smes.yml
```diff
@@ -95,11 +95,11 @@
 - type: entity
   parent: BaseSMES
   id: SMESBasic
-  suffix: Basic, 8MW
+  suffix: Basic, 8MW # CD: 16MW
```
dffdff2423, line 98: Probably should not be a comment.

dffdff2423 (changes requested):

Resources/Prototypes/Entities/Structures/Power/smes.yml
LankLTE: Yeah fair

PR #257, opened 2024-05-12
Increase ore bag capacity

MozarteanChaos:
## About the PR
Ore bags can fit four stacks of ore now.
It's a webedit, but it should work, right? :clueless:

## Why / Balance
Salvaging with the previous ore bag capacity was kind of painful - 175 was only enough room for slightly under three stacks. Now it should be less painful.

## Technical details
This changes two numbers - it sets the capacity to 240, and the size to 241, in line with the old capacity and size (175 and 176 respectively).

## Media

## Breaking changes
Shouldn't break anything, I'd hope.

**Changelog**

:cl:
- tweak: Ore bags can now fit considerably more ore.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/Salvage/ore_bag.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Salvage/ore_bag.yml
@@ -14,9 +14,9 @@
     slots:
     - belt
   - type: Item
-    size: 176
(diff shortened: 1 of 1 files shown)


LankLTE:
Would have no issue with this being raised more- I honestly don't see ore bag balance as that pressing. No idea how good it is with gridinv however.

PR #259, opened 2024-05-12
Skip loading records from cyborgs.

dffdff2423:
## About the PR

Skips loading cyborg records. I will probably add proper cyborg records at some
point. I just don't want to do it until records are in a more stable state. I
also want whatever "maintenance records" I create to apply to synth characters
as well.

## Why / Balance

Graham is not a 90 year old female moth.

## Technical details

We just add an if statement that skips borgs when records are loaded into the game.


LankLTE:
Who are you to say Graham isn't a 90 year old female moth.

PR #277, opened 2024-05-29
Updates Aspid

Aexxie:
adds robotics console to aspid


Nairodian:
That file diff is insane, map moment.

Nairodian (approved):
I'm going to trust this does as it says and that's it because I literally can not easily check the difference.

dffdff2423:
The test failure is related to the map changes.

dffdff2423:
We should probably move aspid to our namespace as upstream is not maintaining it anymore (like we did with gemini).

LankLTE:
Closing this, ask for it to be opened if you have intent to ever finish it.

PR #281, opened 2024-06-04
Port open bag and belt binds from wizden.

dffdff2423:
## About the PR

The keybinds just never got added for some reason.

## Why / Balance

It would make it easier to not need to keep your bag open at all times which is
just good for QOL.

## Technical details

I just copy-pasted the code from upstream and adapted it for listinv.

## Media

## Breaking changes

None

**Changelog**

:cl:
- add: The open backback and open belt keybinds have been ported from wizden.


dffdff2423:
The test failure is a test that is flaky upstream. It should be fixed next time we merge upstream.

LankLTE:
This was never ported as they were only added in the first place to accommodate singleton. I don't have an issue with them being here but I figured I could / should justify that.

dffdff2423:
No worries, the reason why I was a bit confused was that the code worked with basically no modification. I figured that the reason it was not included initially was because it would have required a bit of work to do and the main reason they were added was because of the singleton issue.

PR #284, opened 2024-06-05
Rollerskate improvements

Just-a-Unity-Dev:
## About the PR

This PR visually and statistically improves how Skates work.

## Why / Balance

I didn't like how they worked.

## Technical details

This lowers the speed damage by 0.05. This also modifies one line of upstream code, to use `TryParalyze` instead of `TryStun`. This was changed so that wearers of the skates would actually fall, and not just stand in the middle of the hallway like an idiot if they accidentally bumped into a table or an airlock.

## Media

**Changelog**
:cl:
- fix: Fixed fun!

### DIFF
--- a/Content.Shared/Clothing/Components/SkatesComponent.cs
+++ b/Content.Shared/Clothing/Components/SkatesComponent.cs
@@ -56,7 +56,7 @@ public sealed partial class SkatesComponent : Component
     /// Defaults for MinimumSpeed, StunSeconds, DamageCooldown and SpeedDamage.
     /// </summary>
     [ViewVariables]
-    public float DefaultMinimumSpeed = 23f;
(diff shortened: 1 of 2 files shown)


LankLTE:
Please give actual justification, not just “I don’t like it”. *Why* do the skates need to be faster and do less damage?
Also, you didn’t lower the damage *by* 0.05, you lowered it *to* 0.05, which I’m not a fan of. There should be consequences for flinging yourself into walls.

Just-a-Unity-Dev:
I don't think the skates should apply enough damage to send you into crit, unless you *really* try to, and with how sensitive the skates are for speed, I think it's a fair trade. As for the speed buff that's some random change I put in for no reason - if it's necessary I can revert it

PR #288, opened 2024-06-08
Off-duty Staff Title

Nairodian:
## About the PR
Adds a passenger alternate title, "Off-duty Staff"

## Why / Balance
Flavor title while still being distinct from passenger.

## Known issues
~~The PDA won't show up when equipped, like, I spent an hour looking through YAML and brute forcing fixes, I genuinely couldn't figure it out. Delta-V has the same issue for their custom PDAs too.~~ Fixed.

## Media
image

**Changelog**
Passengers have been given an alternate title in their loadouts to show up as "Off-duty Staff"


Resources/Prototypes/_CD/Entities/Objects/Devices/pda.yml
```diff
@@ -101,3 +102,17 @@
       state: pda-roboticist
     - type: RenameId
       value: cd-job-title-roboticist
+
+- type: entity
+  id: OffDutyPDA
+  parent: BasePDA
```
dffdff2423, line 108: If everything here is the same as the passenger PDA it might be better to just inherit from it instead of BasePDA.

Nairodian: See no other PDA does this so I wasn't sure if I wanted to vary at all. (Also this might make more since I'm going to add the sprites now.)

dffdff2423: Yeah, if you add sprites there is basically no reason to do that.

PR #291, opened 2024-06-10
Provide a default stack size if none is provided

dffdff2423:
## About the PR

This fixes the issue where the game would not correctly calculate
stack sizes because the stack size was null.

## Why / Balance

Bigfix

## Technical details

The ItemSize field was not being set in any of the stack YAML files. This
caused a null check to fail in ItemSystem which made the stack sizes not
update.

## Media

## Breaking changes

None

**Changelog**
N/A


dffdff2423:
From the test failures I don't think this was the correct fix.

PR #299, opened 2024-06-13
Add Hailing Frequency and "encryption key"

dffdff2423:
## About the PR

Adds a frequency that can be used for hailing ships. This was added to the command headset.

## Why / Balance

Giving foreign ships access to station radio just feels kind of wrong, and it fits in
with the SOP changes we are doing. The CC headset was intentionally omitted
from having it because it can broadcast cross-map. There is not really a good
solution to make it not cross-map because the only option to not tie it to a
server is `longRange: true`. It would not make sense to tie it to a server
because it is supposed to be "unencrypted" (ie. Just sending radio off and
hoping people hear)

## Technical details

Adds the channel and key.

## Media

hailing

## Breaking changes

**Changelog**

:cl:
- add: The Hailing radio frequency has been added for ship-to-ship contact. Command Headsets start with a "high power transponder" that can be used to access this frequency. In addition, a "long range headset" has been added that ship based ghost roles start with.


Nairodian (approved):
Looks good to me.

dffdff2423:
Each of the visiting ship ghost roles should have access to this encryption key. I am not sure of the best way to do that.

dffdff2423:
I handed out a new "long range headset" to the various ship ghost roles. I ether gave them the extra headset (in the case of nanotrasen employees) or overwrote their existing headset. I justify this "lore wise" because non-NT employees would not have access to encryption keys for NT stations and it is understandable that a shuttle might have an emergency hailing headset somewhere.

PR #308, opened 2024-06-19
Revert part of 'Make station components use StationPostInitEvent' 

LankLTE:
This is to, hopefully, fix arrivals being paused. Entirely untested, but that's why we're a playtest baby.


dffdff2423 (approved):
Probably fine for now, but I don't think we should keep this long-term.

PR #310, opened 2024-06-22
Hopefully Fix Arrivals Pausing

LankLTE:
Now arrivals spawns will unpause the map they're on when used. This way people won't spawn on paused maps.


dffdff2423 (commented):
This is kinda cursed. Might want to revert the other one given that it did not work so you don't need to deal with the merge conflicts.

PR #313, opened 2024-06-25
Map patching system.

moonheart08:
## About the PR
Implements a mechanism to apply patches to maps (currently, only adding new entities) without modifying the map file itself, only a patchfile.

## Why / Balance
Maps aren't really something you can resolve conflicts in right now.

## Technical details
Provides a verb to add an object to the patch set, and commands to print the set, save the set, and clear the set. Intended for local use, I didn't implement a per-player patchset. Saves into the server's data directory.

## Media

## I have tested my code
- [x] I will eat my hat if it's obviously broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ We are not currently accepting translations of the game on our main repository. 
 
 ## License
 
-Almost code for the content repository is licensed under [MIT](https://github.com/space-wizards/space-station-14/blob/master/LICENSE.TXT). Some code is under the [MPL v2.0](https://www.mozilla.org/en-US/MPL/) and must remain open source.
+Almost code for the content repository is licensed under [MIT](https://github.com/space-wizards/space-station-14/blob/master/LICENSE.TXT). Some code is instead under the [MPL v2.0](https://www.mozilla.org/en-US/MPL/) and must remain open source (and for some, unincumbered by transformative licenses like the GPL as per the Incompatible With Secondary Licenses clauses). All files under the MPL are clearly marked with a file header.
 
 Most assets are licensed under [CC-BY-SA 3.0](https://creativecommons.org/licenses/by-sa/3.0/) unless stated otherwise. Assets have their license and the copyright in the metadata file. [Example](https://github.com/space-wizards/space-station-14/blob/master/Resources/Textures/Objects/Tools/crowbar.rsi/meta.json).
 
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Our [docs site](https://docs.spacestation14.io/) has documentation on SS14s cont
 
 ## Contributing
 
-We are happy to accept contributions from anybody. Get in Discord if you want to help. We've got a [list of issues](https://github.com/space-wizards/space-station-14-content/issues) that need to be done and anybody can pick them up. Don't be afraid to ask for help either!  
+We are happy to accept contributions from anybody. Get in Discord if you want to help. We've got a [list of issues](https://github.com/space-wizards/space-station-14-content/issues) that need to be done and anybody can pick them up. Don't be afraid to ask for help either!
 Just make sure your changes and pull requests are in accordance with the [contribution guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
 
 We are not currently accepting translations of the game on our main repository. If you would like to translate the game into another language consider creating a fork or contributing to a fork.
@@ -31,7 +31,7 @@ We are not currently accepting translations of the game on our main repository. 
 
 ## License
 
-All code for the content repository is licensed under [MIT](https://github.com/space-wizards/space-station-14/blob/master/LICENSE.TXT).
+Almost code for the content repository is licensed under [MIT](https://github.com/space-wizards/space-station-14/blob/master/LICENSE.TXT). Some code is under the [MPL v2.0](https://www.mozilla.org/en-US/MPL/) and must remain open source.
 
 Most assets are licensed under [CC-BY-SA 3.0](https://creativecommons.org/licenses/by-sa/3.0/) unless stated otherwise. Assets have their license and the copyright in the metadata file. [Example](https://github.com/space-wizards/space-station-14/blob/master/Resources/Textures/Objects/Tools/crowbar.rsi/meta.json).
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ We are not currently accepting translations of the game on our main repository. 
 
 ## License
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -31,7 +31,7 @@ We are not currently accepting translations of the game on our main repository.
 
 ## License
 
-All code for the content repository is licensed under [MIT](https://github.com/space-wizards/space-station-14/blob/master/LICENSE.TXT).
+Almost code for the content repository is licensed under [MIT](https://github.com/space-wizards/space-station-14/blob/master/LICENSE.TXT). Some code is under the [MPL v2.0](https://www.mozilla.org/en-US/MPL/) and must remain open source.
```
dffdff2423, line 34: I think you need to further clarify this. The "Incompatible With Secondary Licenses" clause of the MPL is not that commonly used and it took me around 40 minutes of googling to understand. I would also specify that "files under the MPL are clearly marked with a comment at the start of the file" because "some code is under the MPL" might lead to a bit of confusion.

dffdff2423 (changes requested):

dffdff2423:
I assume that Lank has already given you pre-approval to use the MPL with Exhibit B clause and he obviously has the final say on it. However, I would strongly encourage you to drop the Exhibit B clause. I think it kind of goes against the philosophy of an open source game to intentionally restrict other people who use an open source license from using your code.

moonheart08:
> I assume that Lank has already given you pre-approval to use the MPL with Exhibit B clause and he obviously has the final say on it. However, I would strongly encourage you to drop the Exhibit B clause. I think it kind of goes against the philosophy of an open source game to intentionally restrict other people who use an open source license from using your code.

This is fundamentally what the GPL does to me when others use my code in that fashion. The Exhibit B clause restricts a very specific type of license (transformative licenses) and prevents them from doing so.

dffdff2423:
> This is fundamentally what the GPL does to me when others use my code in that fashion. The Exhibit B clause restricts a very specific type of license (transformative licenses) and prevents them from doing so.

The MPL already forces you to duel license your code if you are licensing it under the GPL. According to both [Mozilla's FAQ](https://www.mozilla.org/en-US/MPL/2.0/FAQ/), and the [FSF's GPL compatibility list](https://web.archive.org/web/20240622042911/https://www.gnu.org/licenses/license-list.en.html#MPL-2.0) [^1] if you want to distribute a "larger work" you must duel license each file under *both* licences. That means that any changes they make to your file must remain under the MPL. From my read of it the only thing that Exhibit B does is make it legally incomparable with GPL code so you can't combine it to create a "larger work". Exhibit B does not stop the user from combining the license with a proprietary license and only releasing the changes to your files (which I know is something that you explicitly want to allow). 


This is probably a stupid take. However, I don't see how letting GPL users use your code is any different than letting codebases with hidden content use your code. It is just that the hidden content is out in the open under the GPL. The reason why I think that the GPL is "more alright" in my eyes is because the GPL imposes stronger copyleft onto the code (which I see as protecting the openness of the code), compared to the MPL's Exhibit B which is mostly a "I don't want GPL users to touch my code".  

Again, this is just my take. I don't really care as much as it seems from this wall of text. I mainly just want to understand your point of view.

[^1]: The Gnu website is currently having some issues, I used archive.org to view that link.

dffdff2423 (commented):
I just tested it out in game. Works nicely. I just have a few small issues. I would not say most of these are blockers, but they would be nice QOL to have.

- If you add the same entity to the patchset twice it creates two lines in the patchset.
- It would be nice to have the ability to remove a single entity from the patchset via a verb.
- I am not entirely sure how to load a patchset into the map editor. This is probably just a skill issue on my part.
- I feel like this is an action that you would be using a ton, do you think it would be a good idea to move it to below VV so that mappers have easy access to it? 

Could you also write a quick guide for how to use it in the CDMapPatches directory that we can link to people who are not coders?

moonheart08:
> I just tested it out in game. Works nicely. I just have a few small issues. I would not say most of these are blockers, but they would be nice QOL to have.
> 
>     * If you add the same entity to the patchset twice it creates two lines in the patchset.
> 
>     * It would be nice to have the ability to remove a single entity from the patchset via a verb.
> 
>     * I am not entirely sure how to load a patchset into the map editor. This is probably just a skill issue on my part.
> 
>     * I feel like this is an action that you would be using a ton, do you think it would be a good idea to move it to below VV so that mappers have easy access to it?
> 
> 
> Could you also write a quick guide for how to use it in the CDMapPatches directory that we can link to people who are not coders?

Will look over this later. Most of these are things I can fix, though entity removal is harder to do conflict-less than it sounds. (Also, won't be moving it below VV, it is not frequent enough ingame to justify that.)

moonheart08:
> > This is fundamentally what the GPL does to me when others use my code in that fashion. The Exhibit B clause restricts a very specific type of license (transformative licenses) and prevents them from doing so.
> 
> The MPL already forces you to duel license your code if you are licensing it under the GPL. According to both [Mozilla's FAQ](https://www.mozilla.org/en-US/MPL/2.0/FAQ/), and the [FSF's GPL compatibility list](https://web.archive.org/web/20240622042911/https://www.gnu.org/licenses/license-list.en.html#MPL-2.0) [1](#user-content-fn-1-13461062dc7db4fee68ad22f0f4430c6) if you want to distribute a "larger work" you must duel license each file under _both_ licences. That means that any changes they make to your file must remain under the MPL. From my read of it the only thing that Exhibit B does is make it legally incomparable with GPL code so you can't combine it to create a "larger work". Exhibit B does not stop the user from combining the license with a proprietary license and only releasing the changes to your files (which I know is something that you explicitly want to allow).
> 
> This is probably a stupid take. However, I don't see how letting GPL users use your code is any different than letting codebases with hidden content use your code. It is just that the hidden content is out in the open under the GPL. The reason why I think that the GPL is "more alright" in my eyes is because the GPL imposes stronger copyleft onto the code (which I see as protecting the openness of the code), compared to the MPL's Exhibit B which is mostly a "I don't want GPL users to touch my code".
> 
> Again, this is just my take. I don't really care as much as it seems from this wall of text. I mainly just want to understand your point of view.
> ## Footnotes
> 
>     1. The Gnu website is currently having some issues, I used archive.org to view that link. [↩](#user-content-fnref-1-13461062dc7db4fee68ad22f0f4430c6)

In short, it's because this community doesn't actually know how the GPL works, and will start a firefight claiming the dual-licensed code is "only GPL" when it's not. Hence being willing to remove it if Lank asks.

moonheart08:
Spurious test failure.

dffdff2423:
> entity removal is harder to do conflict-less than it sounds. 

It is not that bit of a deal. I don't expect these to get too large to just edit out manually in their current form.

dffdff2423 (approved):

moonheart08:
> > entity removal is harder to do conflict-less than it sounds.
> 
> It is not that bit of a deal. I don't expect these to get too large to just edit out manually in their current form.

The intent is they stay small, yea.
Something like room patching would likely be an addition to this system using map files and some heuristics for merging, so separate PR (as it'd likely double the length of this PR as is.)

LankLTE:
I'm not a huge fan of outright restricting other licenses that projects can use. I do agree some people might not understand how it works, but with a disclaimer in the file that any re license must also be dual-licensed as MPL (if that is how it works) that should be mostly mitigated, or ignored if they're complaining about something they're outright wrong about.

PR #314, opened 2024-06-26
Mobster Accent Trait

Nairodian:
## About the PR
Adds the mobster accent as a trait.

## Why / Balance
People asked for it enough in ahelp.

**Changelog**
You can now select the mobster accent as a trait in character customization.


LankLTE:
caipche?

PR #315, opened 2024-06-26
Make Oasis 55 pop

LankLTE:
I hate this map. Nairod wouldn't let me derotate it but, at the very least it is really big and probably warrants a high starting pop.


dffdff2423:
Seems fine, upstream Oasis has the same number of players as Fland (which is disabled on CD).

PR #319, opened 2024-06-29
Salvage and Senior role timers

Nairodian:
## About the PR
Adds a two hour overall role timer to salvage.
Adds a three hour overall role timer to all senior roles and warden.

## Why / Balance
Salvage is for grief protection.
Warden and senior roles are kind of arbitrary but it felt right because they can be treated as 2ICs, as well as the fact that only one slot exists for each of them so ensuring griefers don't get those spots is good.

**Changelog**
Salvage now has a two hour game time role timer.
Warden and all senior roles now have a three hour game time role timer.


LankLTE (approved):
1984

PR #320, opened 2024-07-01
Add SOP to guidebook

dffdff2423:
## About the PR

Added Nairod's SOP to the guidebook.


Resources/ServerInfo/Guidebook/_CD/sop/core.xml
```diff
@@ -0,0 +1,61 @@
...
+  Used when determining secure trespass from regular trespass. This is a list of most if not every room that is considered secure. For edge cases the Head of Security and Captain can agree to retroactively declare a room secure.
+
+  - AI core and control rooms.
+  - Armory.
+  - Gravity generator room.
+  - Head of staff's personal rooms.
+  - Main atmospherics room.
+  - Server rooms, including camera, crew monitoring, research, and telecommunications.
+  - Station wide power generation room.
+  - Vault.
```
dffdff2423, line 17: Should bridge be a secure area?

Resources/ServerInfo/Guidebook/_CD/sop/core.xml
```diff
@@ -0,0 +1,61 @@
+<Document>
+  # GENERAL STANDARD OPERATING PROCEDURES
+
+  The standard procedures followed on any uniform NT station. The Captain can edit and make additions to their station's general operation procedures (excluding Contraband) whenever they please as long as a copy of the changes is forwarded to each station bound fax machine. Anyone in the station's sector is required to follow this document and their related work SOP.
```
dffdff2423, line 4: Faxing a copy to every single fax would be a massive pain.

Resources/ServerInfo/Guidebook/_CD/sop/core.xml
Nairodian: Hm- It's a toss up question. Because they can't really _do_ much without access, but it's still a place with a ton of secure terminals.

Resources/ServerInfo/Guidebook/_CD/sop/core.xml
Nairodian: I don't know how else to be certain it's properly relayed for new-arrivals.

Resources/Prototypes/Guidebook/rules.yml
```diff
@@ -1,15 +1,20 @@
...
+#  text: "/ServerInfo/Guidebook/ServerRules/DefaultRules.xml"
 
 - type: guideEntry
   id: CosmaticDriftRules
   name: guide-entry-rules-cd
   ruleEntry: true
   priority: 0
   text: "/ServerInfo/Guidebook/ServerRules/CosmaticDriftRules.xml"
+  children:
+  - CDSop
```
dffdff2423, line 17: Right now I have it under Rules. Do you think that is a good place to put it?

Resources/ServerInfo/Guidebook/_CD/sop/core.xml
dffdff2423: Yeah, I really can't think of a better solution ether.

Resources/ServerInfo/Guidebook/_CD/sop/core.xml
dffdff2423: Intuitively, I would say it is. However, I do think that it is no more risky than the security break room.

Resources/Prototypes/Guidebook/rules.yml
Nairodian: Yes and no. It's rules to security and command, but it could cause confusion to anyone else it isn't part of the rules to (the majority), it is also an in-character document, which rules imply it is not. So perhaps we need a section for "In-character documentation" and place it under that.

dffdff2423: Good point, I will move it tomorrow

PR #322, opened 2024-07-04
Reduced the size for jaws of life so it can fit in normal toolbelts.

z9ctn4fn:
## About the PR

The size of jaws of life is now 20 so that it can fit in normal toolbelts.

## Why / Balance

It's kinda stupid having to keep a jaws of life in your backpack when you have a perfectly functional toolbelt.

:cl:
- tweak: Changed the jaws of life size from 50 to 20.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Tools/jaws_of_life.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/jaws_of_life.yml
@@ -11,7 +11,7 @@
     sprite: Objects/Tools/jaws_of_life.rsi
     state: jaws_pry
   - type: Item
-    size: 50
+    size: 20
   - type: Clothing
     sprite: Objects/Tools/jaws_of_life.rsi
     quickEquip: false


dffdff2423:
Jaws of Life are massive IRL. There is zero chance that they can just fit on a normal toolbelt. In my mind the CE's toolbelt is doing something special to make them fit.

z9ctn4fn:
"Jaws of life compressed through the magic of science"

Nairodian:
It works this way Upstream so I don't mind.
image

dffdff2423:
Yeah, they are clearly some kind of mini jaws of life anyways because they can already fit in a backpack and are able to cut wires.

LankLTE:
I've outright said I don't like the way that upstream does this. The jaws of life are intended to be big and bulky, and they're supposed to be carried around in your backpack as a downside. The most I'd be willing to put them down to is like, 35, to be in-line with the syndicate JoL (which are also intentionally lighter, but given they never make an appearance, w/e).

PR #336, opened 2024-07-28
Revert "Revert "Extend Round End Time""

dffdff2423:
Reverts cosmatic-drift-14/cosmatic-drift#114

I asked in axolotl's dev chat a long time ago how they extend the round end and YuNii said they just change this CVAR. It must have been broken when we tried to change it. The only difference is Axolotl changes it in the config instead of the repo. If this does not work I don't know what to say.


LankLTE:
I'm like 99% sure this is still gonna be broken but whatever why not I love 4 revert PRs

Issue #347, opened 2024-08-04
Mechanically Enforce the two departments + service rule


dffdff2423:
It should not be possible to ready up or late join as a job that is not within your "two departments"

MelliferousMae:
Looking to flesh this out.

List of departments for the purposes of this rule:
High Command (HoP counts as high command and not civilian, probably)
Security
Cargo
Engineering
Medical
Science
Civilian (Can have as many jobs as wanted if there is only 1 other department, can only have 1 "side" job if there are 2 other departments)

Silicon is irrelevant, and station specific and command aren't real departments.

Are there any exceptions to this behavior (I'unno, Prisoner maybe)? Should there be any way built in to be exempted (allowed to pick more job prefs/latejoin as a different job)? Should jobs beyond the limit be greyed out in the character menu so you can't set them, or should you be able to edit freely and then to save they have to validate?

dffdff2423:
Honestly, I am not exactly sure the best way to implement this. There are quite a few ways I can think of doing it. I don't think it needs to be something very high-effort, just enough to stop somebody who forgot the rule exists. People who break it intentionally are fairly obvious. One thing to consider is that players regularly rotate which slots they have selected, so you probably need some way to "commit" the selected departments. 

> List of departments for the purposes of this rule:
> High Command (HoP counts as high command and not civilian, probably)

Yep, that is correct. High command is just HoP and Cap.

> Security
> Cargo
> Engineering
> Medical
> Science
> Civilian (Can have as many jobs as wanted if there is only 1 other department, can only have 1 "side" job if there are 2 other departments)
>
All looks correct.

> Silicon is irrelevant, and station specific and command aren't real departments.
> 
> Are there any exceptions to this behavior (I'unno, Prisoner maybe)? 
You are probably right about prisoner, although not sure what @Nairodian thinks. Passenger would also be another exception.

> Should there be any way built in to be exempted (allowed to pick more job prefs/latejoin as a different job)?
I think it would be nice if it gave you a scary popup and allowed you to violate it. Possibly, with some kind of admin log.

> Should jobs beyond the limit be greyed out in the character menu so you can't set them, or should you be able to edit freely and then to save they have to validate?
I will let you decide what you think is best.

PR #350, opened 2024-08-04
Reorganize roles and add warning about the two job rule

dffdff2423:
## About the PR

- **Recategorize jobs to fit their use on CD**
- **Warning about job rule for HumanoidProfileEditor**

## Why / Balance

This is just a ""temporary"" measure until somebody enforces this mechanically. See #347 

## Media

image

**Changelog**

:cl:
- add: Reorganize the character editor to make it clear what jobs are in what department for the "two department + service" rule

### DIFF
--- a/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml
+++ b/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml
@@ -119,6 +119,7 @@
                 </BoxContainer>
                 <BoxContainer Orientation="Vertical">
                     <!-- Jobs -->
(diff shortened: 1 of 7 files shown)


Nairodian:
How does this look on the crew manifest, you know?

dffdff2423:
image

Nairodian:
Hm, I realize it's not too bad either way. Being able to quickly take a glance at the manifest and get a list of all the command members was nice, but perhaps just ask on the Discord what others think.

LankLTE:
I'm with Nairod that it's nice to just be able to look and see every command member

dffdff2423:
Then I think re-adding Command but then also adding "High Command" might be the best approach.

dffdff2423:
Image of current state: 
image

PR #356, opened 2024-08-08
Disable admin rank in ahelp window

dffdff2423:
Disables showing the admin rank in the ahelp window per admin consensus.


dffdff2423:
This is a borked test upstream

PR #360, opened 2024-08-10
make tied shuttle votes pass

LankLTE:
biggest mald PR of my life (I don't really think we should continue a round where 50% of the server wants to leave)


dffdff2423 (approved):

Ariphylor:
image_2024-08-10_155840621

PR #363, opened 2024-08-10
Redo Figurine Lines

Nairodian:
## About the PR
Redoes a majority of the new figurine lines to fit the CD environment and add a couple inside jokes.

## Why / Balance
A lot of the original ones just didn't suit CD.

## Technical details
(I did not mark every single difference, instead just labeling the top of the file with "Heavily edited by CD." I can adjust this if Lank or Aquif would like.)

**Changelog**
Most figurines have had their voice lines changed or had lines added.


dffdff2423 (approved):
I love these.

LankLTE (changes requested):
1984

PR #366, opened 2024-08-17
(hopefully) make the event scheduler less chaotic

LankLTE:
I'm not 100% sure how exactly this works, but I *believe* this should result in less events spawning, more similar to how it was before the refactor.


dffdff2423:
I did not test a full 4 hour round, but I tried it for a bit and I think this looks fine, we are going to need to keep tweaking these numbers anyways.

dffdff2423 (approved):

PR #369, opened 2024-08-18
Nerf event scheduler (again) and gas leak

dffdff2423:
## About the PR

title

## Why / Balance

This is just a minor nerf mainly because of how many of the past few rounds
have been hijacked by events. It might need to be nerfed even more but I think
slowly toning it down is the best approach.

## Technical details

YAML

## Media

N/A

## Requirements

- [X] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion

**Changelog**

:cl:
- tweak: Gas leaks should be smaller and less frequent.


Resources/Prototypes/GameRules/events.yml
```diff
@@ -259,7 +259,7 @@
     startAudio:
       path: /Audio/Announcements/attention.ogg
     endAnnouncement: station-event-gas-leak-end-announcement
-    weight: 10
+    weight: 5 # CD: Nerf weight, please revert if upstream ever makes "dangerous" gasses more rare.
```
dffdff2423, line 262: I am not happy with decreasing the weight for the reasons we have talked about previously but I can't really think of a better solution other then making dangerous gasses more rare which I don't have the time to make an upstream PR for at the moment.

PR #374, opened 2024-08-19
More Map Patches

Nairodian:
## About the PR
Adds a map patch for the rest of the remaining non-CD maps that didn't have one. (Excludes Reach, because Reach, don't think I need to explain that one.)
##
**Box**
- Adds a LnF locker in the command conference room above HoP.

**Cog**
- Adds a LnF locker in HoP's bedroom.

**Oasis**
- Adds a LnF locker in the command conference room.
- Adds an employment records computer in every Head room excluding HoP and RD. (Space issues.)

**Omega**
- Adds a LnF locker in the bridge.
- Adds a an employment records computer in the bridge.
- Adds a medical records computer by the surgery room.

**Packed**
- Adds a LnF locker in the bridge.

## Why / Balance
Lost and Found lackage.

**Changelog**
Every map should have a Lost and Found locker now, plus records computers where relevant.

### DIFF
--- a/Resources/CDMapPatches/box.yml
+++ b/Resources/CDMapPatches/box.yml
@@ -1,4 +1,3 @@
 - !type:CDSpawnEntityMapPatch
   worldPosition: -5.999672,-16.99885
   id: LockerLostAndFound
-...
--- a/Resources/CDMapPatches/box.yml
+++ b/Resources/CDMapPatches/box.yml
@@ -1,0 +1,4 @@
+- !type:CDSpawnEntityMapPatch
+  worldPosition: -5.999672,-16.99885
+  id: LockerLostAndFound
+...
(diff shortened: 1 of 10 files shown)


Resources/CDMapPatches/box.yml
```diff
@@ -0,0 +1,4 @@
+- !type:CDSpawnEntityMapPatch
+  worldPosition: -5.999672,-16.99885
+  id: LockerLostAndFound
+...
```
dffdff2423, line 4: You should delete these `...` lines. They are an artifact of the YAML library SS14 uses.

dffdff2423 (changes requested):

Resources/CDMapPatches/box.yml
Nairodian: Should be fixed now.

dffdff2423 (approved):

PR #377, opened 2024-08-21
Remove Mantle Timers

Nairodian:
## About the PR
Removes the time-gate on all head mantles.

## Why / Balance
Doesn't fit with our style, you could get them on the Terminal round-start anyways.

**Changelog**
Mantles are no longer time locked in loadouts.


LankLTE (approved):
I was gonna ask you to do this and forgot

PR #399, opened 2024-09-07
Dogtags

### LINKED ISSUE #22: [Feature Request] Dogtags
## Description
The hope is this would port of CM's dogtags but alter them to be more RP focused than they are usually. Here is the hope.

Wearable on the neck slot. Probably nothing else.

You have an option once (after you do this it disappears and can't happen again) where you can change the item's description slightly, probably around a forty-fifty character limit where you can put whatever information you would like to. This action could be labeled as "Engrave Dogtags" or something similar. This is so we aren't limiting information to just character name, though names should likely be included somewhere.

While it isn't entirely needed since this is primarily a flavor item for RP, I wouldn't be opposed to setting up the snapping off a dogtag system for dead bodies. The dogtags once engraved get a new action probably called "snap dogtags" which will give two new dogtags, with a different sprite. (A single dogtag instead of the standard two) Both of which would have the same description as what they were engraved with.

## Technical Details
In order to do it properly, I'd like to wait for [loadouts](https://github.com/space-wizards/space-station-14/pull/17517) to be merged here so people can spawn with them if they want to use them and then we won't have to map them.

We'd likely just take the sprites from CM and credit them, but I'm more than open to us just making our own.

LankLTE:
Can be engraved once to edit the description, available as a trinket. Sprite is currently just CM's, though I'd prefer a custom one that fits the artstyle better. 

Closes #22


Content.Server/_CD/Engraving/EngraveableComponent.cs
```diff
@@ -0,0 +1,39 @@
+using Robust.Shared.GameStates;
+using Robust.Shared.Utility;
+
+namespace Content.Server._CD.Engraving;
+
+/// <summary>
+/// Allows an entity to be prayed on in the context menu
```
dffdff2423, line 7: Copy paste?

Content.Server/_CD/Engraving/EngraveableSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+using Content.Shared.Verbs;
+using Robust.Shared.Player;
+
+namespace Content.Server._CD.Engraving;
+/// <summary>
+/// System to handle subtle messages and praying
+/// </summary>
+/// <remarks>
+/// Rain is a professional developer and this did not take 2 PRs to fix subtle messages
+/// </remarks>
```
dffdff2423, line 17: Copy paste?

dffdff2423 (dismissed):
Looks fine other then a few typos.

Content.Server/_CD/Engraving/EngraveableSystem.cs
LankLTE: Let's pretend like this never happened

PR #401, opened 2024-09-09
Make shuttle votes use game time

dffdff2423:
## About the PR

Fixed shuttle votes not using game time and the extremely cursed way we prevented people from recalling the shuttle.


LankLTE:
Instead of changing the extension time, change the inital vote time to be ~15 minutes lower. This ensures between the vote, 10 minute timer, then transit / arrivals time the round should end approx. on the hour, which was the intention.

dffdff2423:
I think the extension time should also be reduced somewhat. 45 minutes often feels too long. I agree with you that lowering the initial time also makes sense.

dffdff2423:
Oh, I was going to change it to the numbers in the CVARs, I guess it does not really matter if they are just going to be overridden anyways.

PR #422, opened 2024-09-21
SOP Books

Nairodian:
## About the PR
Adds an SOP book for each SOP guidebook entry.
Places them rarely in related department lockers and bookcases.

## Why / Balance
Mostly just to draw attention to the SOP, but also gives people confirmation that the SOP is an in-world document.

## Media
image

**Changelog**
New SOP books linking to the SOP guidebook entries will occasionally spawn in their related department lockers.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Books/bookshelf.yml
+++ b/Resources/Prototypes/Catalog/Fills/Books/bookshelf.yml
@@ -23,6 +23,15 @@
       - id: BookSpaceEncyclopedia
       - id: BookSpaceLaw
       - id: BookTheBookOfControl
+      #CD Addions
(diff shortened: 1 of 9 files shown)


dffdff2423:
I think command lockers should have a 100% chance to have one of the books given that command is expected to be aware and following SOP.

Nairodian:
> I think command lockers should have a 100% chance to have one of the books given that command is expected to be aware and following SOP.

Made all heads have a 100% chance to start with their department SOP, and the command book was given a slightly higher chance to spawn. I just simply don't want there being an overabundance of any given book.

dffdff2423:
Yeah, that works

PR #431, opened 2024-09-28
Hide Round End & Map Vote Counts

LankLTE:
this was kinda always the plan but now it's possible


dffdff2423:
Can you also disable it for map votes if they are not already?

LankLTE:
ok

PR #442, opened 2024-10-21
Add human hair to vulps

dffdff2423:
Sector Umbra backport.
https://github.com/Sector-Umbra/Sector-Umbra/pull/141

Not tested in-game but probably fine.

CL: Vulpkin can now use human hair styles. (Credit to SharkskyRaeJepsen)

I don't think they are in the CD discord.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Customization/Markings/human_hair.yml
+++ b/Resources/Prototypes/Entities/Mobs/Customization/Markings/human_hair.yml
@@ -2,7 +2,7 @@
   id: HumanHairAfro
   bodyPart: Hair
   markingCategory: Hair
-  speciesRestriction: [Human, Dwarf, SlimePerson, Reptilian, Moth]
+  speciesRestriction: [Human, Dwarf, SlimePerson, Reptilian, Moth, Vulpkanin] # Umbra: Vulpkanin Allowed
   sprites:
     - sprite: Mobs/Customization/human_hair.rsi
       state: afro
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Entities/Mobs/Customization/Markings/human_hair.yml
```diff
@@ -2,71 +2,71 @@
   id: HumanHairAfro
   bodyPart: Hair
   markingCategory: Hair
-  speciesRestriction: [Human, Dwarf, SlimePerson, Reptilian, Moth]
+  speciesRestriction: [Human, Dwarf, SlimePerson, Reptilian, Moth, Vulpkanin] # CD: add Vulps and moths
```
LankLTE, line 5: Technically reptilians should be on the header too as their hair is a CD change. Not the end of the world though since it being there at all is an improvement.

LankLTE (commented):
Should also note it'd be best if you could check if there's any that were missed (assuming this is a direct port from umbra) because of upstream merges & the like.

Resources/Prototypes/Entities/Mobs/Customization/Markings/human_hair.yml
dffdff2423: Oh, I did not know that. That is one sed command.

dffdff2423:
I looked through it and I think everything with a `speciesRestriction` is covered

PR #460, opened 2024-11-05
[RFC] Revert low pressure damage

dffdff2423:
## About the PR

Reverts commit 37fffca6155b1b02398baca38677f17a7abddbfe.

## Why / Balance

The major argument for reducing the damage in https://github.com/space-wizards/space-station-14/pull/29478 boils down to "what if the shuttle gets bombed". This is a non-issue on CD given that ideally the shuttle should never be bombed without a very good reason.

In my opinion spacing *should* be a punishing and unfair mechanic. If you are going out into space without a spacesuit you are doing something that should be highly mechanically discouraged. Given that nature of CD being a RP server it is fairly immersion breaking to see John Paramedic going out into space to recover the body. If you blow a hole in the wall of the space station it *should* be a catastrophic event with little counterplay; I think the drama from that would add a lot to a CD round.

Harmony has also reverted low pressure damage and it does not seem to be a dumpster fire based on their github https://github.com/ss14-harmony/space-station-14/pull/137

**Changelog**

aquif:
- Low pressure damage has been restored to it's pre-nerf values.

### DIFF
--- a/Content.IntegrationTests/Tests/Commands/SuicideCommandTests.cs
+++ b/Content.IntegrationTests/Tests/Commands/SuicideCommandTests.cs
@@ -168,7 +168,7 @@ public sealed class SuicideCommandTests
         await pair.CleanReturnAsync();
     }
 
-        /// <summary>
+    /// <summary>
     /// Run the suicide command in the console
     /// Should only ghost the player but not kill them
     /// </summary>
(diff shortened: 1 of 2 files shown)


dffdff2423:
I will fix the test if this is approved.

dffdff2423:
Cherry-picked https://github.com/space-wizards/space-station-14/pull/33030. Do not sqash-merge this PR.

LankLTE:
If I could, I'd say maybe partially revert it to a middleground. Just to play around with a bit & see, the current value is a big jump and might take a lot of people off-guard.

dffdff2423:
I would be fine with that, but I did not see a single person object to this on discord. Spacing rarely comes up in CD as is.

LankLTE:
Personally I'd say put it to 2-3 just so it's not as jarring both for new and existing people, but if you insist 4 I'm fine with that ultimately.

PR #464, opened 2024-11-09
Increase Character Description Examine Range

Nairodian:
## About the PR
Makes it so you can view character descriptions from 8 tiles away instead of 3.

## Why / Balance
I remember RMC doing this and in general, I don't want to have to hug someone to read their typically at least one paragraph long description.

## Media
image

**Changelog**
You can now view character descriptions from a much greater range.


dffdff2423:
I have been intending to do this for a while, I just keep forgetting.

PR #473, opened 2024-11-18
fix non-whitelisted slots not working 

MilonPL:
## About the PR

title

the return was the wrong way around
the connecting player (data.UserId) isn't in the connected players list yet, so theres no need to check their status


dffdff2423 (approved):

MilonPL:
I'm going to fucking NUKE DELETEALLTHENGHOST

PR #480, opened 2024-11-21
admin event preference panel

lunarcomets:
## About the PR

adds an event preference panel that allows admins to see the event preferences of a player's current selected character through the ```eventpreferences <name>``` command

currently only accessible through a command, can be added if needed or if it's wanted via context menu, admin menu, etc.

## Why / Balance

while working on the character panel, i figured that there will be situations that an admin would only want to see a player's event preferences without the additional features that will be present in the character panel

whether or not this gets merged or is needed, the xaml file will be reused for the characterpanel itself

## Technical details

implemented with Eui, a player's character current character profile is fetched and the current active preferences is gotten from the profile every time that the Eui is opened via command or (hopefully) otherwise

the preferences that are listed in the panel are every AntagPrototype that has it's VisiblePreference bool set to True, allowing further preferences that may be added in preference.yml to be listed here automatically without the need of adding anything to the xaml or code as well

a new EventPreferences control was added under in the Content.Client._CD.Admin.UI namespace as well, allowing this portion of the UI to be easily reused

## Media

image

## Requirements

- [X] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion

## Breaking changes

nada

**Changelog**

for admins: added an event preference panel that allows game admins to view the event preferences of a player via the ```eventpreferences``` command


lunarcomets:
something fucked up with the pushing the FUCKK

Resources/Locale/en-US/_CD/admin/admin.ftl
```diff
@@ -2,3 +2,10 @@ admin-player-actions-window-cd-record-purge = Modify Character Records
 cd-actions-admin-modify-records = Modify Character Records
 cd-actions-admin-modify-reset = Reset Records
 cd-actions-admin-modify-del-entry = Delete Entry
+
+cd-eventpreferencepanel-title = Event Preference View
+cd-eventpreferencepanel-player = Player: {$player}
+cd-eventpreferencepanel-character = Character: {$characterName}
+
+cmd-eventpreferences-help = Usage: eventpreferences <name>
+cmd-eventpreferences-desc = Displays a player's current character's event preferences (i.e. antags)
```
dffdff2423, line 11: nitpick: can you add a cd prefix to these locale keys? These are generic enough that I could see upstream adding them.

Content.Server/_CD/Admin/EventPreferencePanelEui.cs
```diff
@@ -0,0 +1,48 @@
...
+        foreach (var antagPrototype in _proto.EnumeratePrototypes<AntagPrototype>())
+        {
+            if(antagPrototype.VisiblePreference)
+                visibleAntagPrototypes.Add(antagPrototype);
+        }
+
+        _visibleAntagPrototypes = visibleAntagPrototypes;
+
+        StateDirty();
+    }
```
dffdff2423, line 47: It should not be needed to manually call this method when the UI is opened. Given that this *should* be constant within a round it can probably be set within the constructor?

Content.Server/_CD/Admin/Commands/EventPreferencesCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+using Content.Server.Preferences.Managers;
+using Content.Shared.Administration;
+using Content.Shared.Preferences;
+using Robust.Server.Player;
+using Robust.Shared.Console;
+
+namespace Content.Server._CD.Admin.Commands;
+
+[AdminCommand(AdminFlags.Admin)]
+public sealed class EventPreferencesCommand : LocalizedCommands
```
dffdff2423, line 13: Toolshed commands are preferred to legacy commands because legacy commands are deprecated upstream, however this is not a blocker on this PR as I don't expect them to be removed any time in the foreseeable future.

Content.Server/_CD/Admin/Commands/EventPreferencesCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    [Dependency] private readonly IPlayerLocator _locator = default!;
+    [Dependency] private readonly IServerPreferencesManager _pref = default!;
+    [Dependency] private readonly EuiManager _euis = default!;
```
dffdff2423, line 17: nitpick: inconsistent name

Content.Client/_CD/Admin/UI/EventPreferencesPanel.xaml.cs
```diff
@@ -0,0 +1,36 @@
...
+            "cd-eventpreferencepanel-character",
+            ("characterName", currentPreferences.Name)
+            );
+    }
+
+    /// <inheritdoc cref="EventPreferences.SetPreferenceList"/>
+    public void SetPreferenceList(HumanoidCharacterProfile userPreferences, IEnumerable<AntagPrototype> listablePreferences)
+    {
+        EventPreferenceControl.SetPreferenceList(userPreferences, listablePreferences);
+    }
```
dffdff2423, line 35: I think these should be condensed into one method because you will always need to call both, and I don't see a situation where they would need to be updated separately.

dffdff2423 (changes requested):
Looks good! Just a few minor things. I reviewed this at 11:40pm, so if anything here is nonsensical let me know.

Resources/Locale/en-US/_CD/admin/admin.ftl
lunarcomets: before the cmd part? because i think how it's currently formatted is how it's being referred to so that it's brought up in the console; if you mean more like cmd-cdeventpreferences-<stuff> then yeah can do

dffdff2423: Oh, right. I am moronic.

Content.Server/_CD/Admin/Commands/EventPreferencesCommand.cs
lunarcomets: there's also the fact that a lot of admins do not know how to work toolshed from when i last remember, so very much won't be getting removed soon, will look into making it a toolshed command in a future pr however

lunarcomets:
no longer dying yippioe

.run/Content Server+Client.run.xml
```diff
@@ -4,4 +4,4 @@
     <toRun name="Content.Server" type="DotNetProject" />
     <method v="2" />
   </configuration>
-</component>
```
dffdff2423, line 7: Random line change that might cause conflicts. Can you revert it?

dffdff2423 (approved):

dffdff2423:
Looking at this again, I would rather preferences not show up if they are not enabled. I think red vs. green could cause a bit of confusion because people commonly associate antags with red. Given that it is a two line change I am just going to make it myself so this gets merged.

lunarcomets:
oh that works too, thank you

PR #482, opened 2024-11-23
PocketVisibleComponent

MilonPL:
## About the PR

marks the component as not obscured even though it's in a pocket

## Why / Balance

feature request dont question it :godo:


dffdff2423:
What exactly was this wanted for?

MilonPL:
> What exactly was this wanted for? I have not been following dev chat that much the past few days.

pouches rmc style except rmc just makes pockets not hide items at all

MilonPL:
this is so going to conflict with upstream #31076 :trollface:

KittenColony:
> What exactly was this wanted for? I have not been following dev chat that much the past few days.

I requested it, the idea is for some items like pouches and holsters to be visible

IE, 'Oh, this is a holster, it doesn't exactly fit *inside* the pocket so it goes on the outside', so people can search and see the holster, as well as things like **I**ndividual **F**irst **A**id **K**its where its easier to know which pocket to immediately remove to search for supplies

dffdff2423:
wtf did Lank fix the label bot and we have the wizden PR sizes that make no sense?

LankLTE:
> wtf did Lank fix the label bot and we have the wizden PR sizes that make no sense?

I just merged upstream and it started doing it again. Pretty sure beforehand I had it disabled intentionally but I might be wrong. 

Anyways as for the PR I'd rather also see whatever it's going to be used for PRed because if we don't want that I don't really want a component that doesn't do anything.

KittenColony:
> Anyways as for the PR I'd rather also see whatever it's going to be used for PRed because if we don't want that I don't really want a component that doesn't do anything.

The immediate first thing that it'll be used in will be Drip 3.3 for the Reagent pouch, and potentially another medical storage pouch, after that it'll be the material and tool pouches for Engineers / Cargo in 4.#, then a holster or sling for Security in 5.#

LankLTE:
PR the other things and say it requires this, I'd rather not merge this unless we actually do want what requires it.

LankLTE:
Will likely close this as pouches aren't a desired feature.

PR #504, opened 2024-12-17
Add antag target preferences.

dffdff2423:
## About the PR

Added preferences so players can opt-in to being antag targets.

## Why / Balance

This is mainly so admins don't need to bwoink players and spoil that they are a target.

## Technical details

I prefixed the locale string for each antag pref with a number. This is because the game sorts them alphabetically. It looks *much* better than the alternative.

**Changelog**

You can now opt-in to being intentionally targeted for general antagonistic activity and round removal under the antag tab in the character creator. Keep in mind not enabling this will not prevent harm from befalling you throughout normal gameplay (our minor crimes rule still applies).


Nairodian (approved):
Assuming this shows up in the new event pref panel. Either way, looks good. I like the shotty solution to the alphabetical order issue.

Issue #357, opened 2024-08-08
It should be possible to vote on round end time while on the shuttle


dffdff2423:
Do a custom vote for how long CC is while on the shift transfer shuttle

MelliferousMae:
This issue looks like it should either be closed or updated as the associated PR was closed for being unwanted.

dffdff2423:
oh thanks

PR #518, opened 2024-12-27
Create vote for the time at CentCom

dffdff2423:
## About the PR

Create vote for the time at CentCom. 

Closes https://github.com/cosmatic-drift-14/cosmatic-drift/issues/357

## Why / Balance

Some rounds there needs to be more time at CC and it is often hard for admins
to gauge if this is wanted from the players. Additionally, CC can drag on for a
bit sometimes and this should hopefully solve this.

## Technical details

Instead of checking for the CVar check for the value set in `RoundEndVoteSystem` (formerly `ShuttleVoteSystem`). We then call a vote after the shuttle is launched.

## Breaking changes

Contains wizden changes.
`ShuttleVoteSystem` -> `RoundEndVoteSystem`

**Changelog**

A vote will be shown during the arrivals terminal to decide how long to spend at the CentCom station. The time will be the weighted average of the votes.

### DIFF
--- a/Content.Server/RoundEnd/RoundEndSystem.cs
+++ b/Content.Server/RoundEnd/RoundEndSystem.cs
@@ -46,7 +46,7 @@ namespace Content.Server.RoundEnd
         [Dependency] private readonly SharedAudioSystem _audio = default!;
         [Dependency] private readonly StationSystem _stationSystem = default!;
 
-        [Dependency] private readonly ShuttleVoteSystem _cdShuttleVoteSystem = default!;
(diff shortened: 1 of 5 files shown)


LankLTE:
We talked about this on discord. I don't remember exactly what the conclusion was but this should probably be changed & a way for admins to delay the round-end mid-CC would be ideal.

dffdff2423:
> We talked about this on discord. I don't remember exactly what the conclusion was but this should probably be changed & a way for admins to delay the round-end mid-CC would be ideal.

yeah, I will dig out what we decided later

dffdff2423:
closing this as it is not wanted. I will make a new PR (or re-open this) when I have time.

PR #529, opened 2025-01-02
make dogtags text predicted

MilonPL:
## About the PR

basically the changes i did in https://github.com/emberfall-14/emberfall/pull/12 anyways so no reason not to PR it here too
moved the examine part to shared and networked the text so that you dont have to wait one second and watch it change from the "nothing engraved" text to "custom message" text in real time


dffdff2423 (approved):
Sorry this took so long to review.

PR #534, opened 2025-01-14
Unrestrict Vox Colors

LankLTE:
Little reason to have this restricted. Will need an announcement saying vox colors will be fucked.


dffdff2423 (approved):
yeah, Avali have the full range of colors.

PR #537, opened 2025-01-20
Russian Accent Trait

LankLTE:
Courtesy of zephyrkushala.


LankLTE:
I've heard concerns upstream of this accent being very difficult to understand for people who actually speak a Cyrillic language due to it simply mixing in Cyrillic characters. Something to keep in mind

dffdff2423:
> I've heard concerns upstream of this accent being very difficult to understand for people who actually speak a Cyrillic language due to it simply mixing in Cyrillic characters. Something to keep in mind

I bet this is also the case for the german accent. I only speak it at a B1 level and it already makes me go insane.

PR #547, opened 2025-02-02
1984 LRP posters

dffdff2423:
## About the PR

Remove posters that can be seen as LRP-ish.
## Why / Balance

Deathsquad recruitment -> We want death squad to be a more mysterious thing.
Lusty Xenomorph Maid -> Why does this exist on a 16+ codebase. Thanks wizden.
No ERP poster -> ERP is not a IC term.
Busty Backdoor Exo Babes -> Why is this in a game that is intended to be rated 16+?! If somebody said the description in-game they would likely get banned.

## Technical details

Disabled them in the random spawner and migrated them out in case any maps use them.

**Changelog**

A few LRP posters have been removed.

### DIFF
--- a/Resources/Prototypes/Entities/Markers/Spawners/Random/posters.yml
+++ b/Resources/Prototypes/Entities/Markers/Spawners/Random/posters.yml
@@ -34,7 +34,7 @@
       - PosterContrabandFreeTonto
       - PosterContrabandAtmosiaDeclarationIndependence
       - PosterContrabandFunPolice
-      - PosterContrabandLustyExomorph
+      # - PosterContrabandLustyExomorph CD: LRP poster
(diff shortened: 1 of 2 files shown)


Afrokada:
might want to remove the "Busty Backdoor Exo Babes" one, too, for the same reason the xenomorph maid poster gets removed.

dffdff2423:
I literally did not know that existed

dffdff2423:
I managed to find the wizden PR removing posters https://github.com/space-wizards/space-station-14/pull/32734#pullrequestreview-2588863815. Do you think we should remove the last one in the wizden PR?

cc @Nairodian

dffdff2423:
That was what I thought as well, but figured I would ask.

PR #552, opened 2025-02-08
Remove Hostile Shuttles

LankLTE:
This must've slipped through.

### DIFF
--- a/Resources/Prototypes/GameRules/unknown_shuttles.yml
+++ b/Resources/Prototypes/GameRules/unknown_shuttles.yml
@@ -27,13 +27,13 @@
     children:
     - id: UnknownShuttleSyndieEvacPod
 
-- type: entityTable
-  id: UnknownShuttlesHostileTable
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/GameRules/unknown_shuttles.yml
```diff
@@ -27,13 +27,13 @@
     children:
     - id: UnknownShuttleSyndieEvacPod
 
-- type: entityTable
-  id: UnknownShuttlesHostileTable
-  table: !type:AllSelector # we need to pass a list of rules, since rules have further restrictions to consider via StationEventComp
-    children:
-    - id: LoneOpsSpawn
-    - id: UnknownShuttleManOWar
-    - id: UnknownShuttleInstigator
```
dffdff2423, line 36: you probably don't need to remove this.

LankLTE: I figured it's better in case it ends up showing up anywhere else, tests will freak out.

PR #561, opened 2025-03-01
Pet Pebble

Nairodian:
## About the PR
Adds Pebble as a station pet, the argocyte from Hemshire's multi-session event.
{This does not include his spawner being mapped.}

## Why / Balance
The consensus was to give the crew a lasting prize if they succeeded in the event.

## Media
image

**Changelog**
Not just yet, probably will save that for when he's actually mapped.


Resources/Prototypes/_CD/Entities/Mobs/NPCs/pets.yml
```diff
@@ -0,0 +1,35 @@
...
+  - type: Tag
+    tags:
+    - CannotSuicide
+    - VimPilot
+  - type: Grammar
+    attributes:
+      proper: true
+      gender: epicene
+  - type: InteractionPopup
+    successChance: 0.7
```
dffdff2423, line 30: This is a massive nitpik, but it is a rock thing. There is not a 30% chance I am going to miss petting a slow moving rock thing.

dffdff2423 (approved):
No issues from me. Probably should have a \@ Mapping Notifications ping (or whatever it is called) if we still use that. (I can't ping it so somebody else will need to merge this)

(on another note might be good to just get rid if it and give map maintainers a pingable role on discord)

Resources/Prototypes/_CD/Entities/Mobs/NPCs/pets.yml
Nairodian: It's not so missing it as it not accepting. The flavor text for a failed interaction is, "You reach out to pet Pebble, but are met with a cold stare."

PR #562, opened 2025-03-03
Template changes

dffdff2423:
Update the PR template and CONTRIBUTING to address the most common reasons PR's get rejected or changes requested.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,7 @@
 <!-- The text between the arrows are comments - they will not be visible on your PR. -->
 
 ## About the PR
-<!-- What did you change in this PR? -->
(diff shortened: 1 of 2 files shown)


LankLTE:
It's been on my todo list to make the PR template far shorter, I don't actually expect people to fill out most of the info there.

dffdff2423:
I removed basically everything other than the "About this PR", "Media", the checkboxes, and the changelog. I think those are really the only things we might care about.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,12 +4,6 @@
 ## About the PR
 <!-- What did you change in this PR? -->
```
LankLTE, line 5: Maybe just add "and why", just because I'd like to know people's reasoning for a PR. Doesn't need a while section though.

LankLTE (changes requested):

PR #565, opened 2025-03-04
Respriting Humans & Reptilian base parts.

KittenColony:
## About the PR
Resprites the Human bodyparts for a less paper-doll appearance, as well as port the most recent Reptilian parts sprites from /tg/

## Why / Balance
The sprites used were old, and this just updates them to be more up-to-date, and be more detailed when more closely inspected.

## Technical details
Copies of the human & reptilian parts.rsi within textures/_CD, adjusted the path within the species .yml files.

## Media
image

## Requirements

**Changelog**
🆑 
Humans and Reptilians now use a more modern design, this should only be noticeable without clothes on / only be visible on close examination.


Nairodian (commented):
Alright as for my actual feedback. I'm generally not a fan of this resprite. It's not all bad, but the main issue is you're making the outline shading extremely difficult to follow. I also don't see a reason to really make things look more _realistic_, every other species follows a slightly cartoonish look. This isn't "better," it's a different art direction, one I don't particularly feel fits for CD.

Human
-Approaches the uncanny valley with that degree of shading, especially on the face.
-The outline shading is way too broken up compared to the other species. Where as every other species looks like it's outline is a solid color all around, my eyes instantly jump to the human's kneecaps.
-Somehow this makes their ass even more defined than before.

Reptile
-The top of their arms stick out way too much because they're using a different outline color. Mean just look at how aggressive the shift is going from arm to hand on the side view.
-Their tail from the front is even more difficult to discern than before because the outline shading uses the same shading as the tail.
-The legs look weird at the side view because the outline suddenly inverts in colors. I get that's supposed to be the other leg but it differs from every other species sprite.

dffdff2423:
After additional discussion we have decided against merging this.

- We do not want the additional maintenance burden of having to maintain a separate art style. This art style may conflict with new markings.
- It suddenly changes the look of a large population of player characters (especially with such a noticeable change like the human sprite) which would go against our core tenant of player expression.
- Every other ss14 server uses these sprites. The only rational you have given is that they "look old" which we do not consider a valid reason considering that a large proportion of other sprites in the game "look old" (see point one).

PR #568, opened 2025-03-05
Item Trait

LankLTE:
Ideally this should have some kind of popup as to why it doesn't work, but I figure this is better than nothing.


Content.Server/_CD/Traits/ScaleItemComponent.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite), DataField]
+    public float? MaxScale;
+
+    /// <summary>
+    /// The minimum scale their player can be set to for this component to itemize them
+    /// Same caveats as MaxScale.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite), DataField]
+    public float? MinScale;
```
dffdff2423, line 31: I don't see the point in having a MinScale that is basically just dead code.

LankLTE: Just in case? idk I wrote this like 4 months ago and didn't check it so if you'd prefer it removed I can do that

PR #569, opened 2025-03-06
Fix Arrivals

LankLTE:
This fix doesn't seem like it *should* work, but to my (limited) testing it did. The map at least loaded and arrivals cycled fine, so.


dffdff2423:
I would also try re-saving the map in-addition to this. I don't think it has been saved in a while.

PR #576, opened 2025-03-17
Make anomalous infections less common

LankLTE:
I think it being 1/3 is probably unintentional, I swear we had it lower originally, but imho this should be rare so 1/20 is fine.


dffdff2423:
We definitely did have it lower. I forgot what the old value was. This is probably fine.

PR #586, opened 2025-04-01
MapVote Exclusion

Aexxie:
## About the PR

Holy shit, I accidentally just deleted everything I wrote. Let's try that again.
- Adds a mapvote exclusion system to prevent recently selected maps from appearing again in mapvote.
- Adds a CVAR, `MapvoteCacheDepth`. This CVAR configures the amount of recent maps that will be excluded from mapvotes. Default is 2.
- Adds `AdvancedRound` to Database. `AdvancedRound` serves as a more detailed version of the `Round` entry. `AdvancedRound` stores the RoundID, ServerID, MapID, and StartDate of each round. This info is required for the rest of the PR.
- Database migrations to support the new entry. These worked fine for me, but I would double check before deploying just in case.
- Changes to `GameTicker` systems to support `AdvancedRound`.
- Changes to different voting systems to make the PR work.

**Changelog**
No changelog required.


Content.Server/_CD/GameTicking/GameTicker.AdvancedRound.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            await _db.AddNewAdvancedRound(server, RoundId, _gameMapManager.GetSelectedMap()!.ID);
+
+            UpdateMapQueue(_gameMapManager.GetSelectedMap()!.ID);
+        });
+    }
+
+    private void UpdateMapQueue(string mapName)
+    {
+        var task = Task.Run(async () =>
```
dffdff2423, line 30: UpdateMapQueue should return a task that you await. If this is the only use I don't see why it needs to be a separate method.

Also, not sure if I am right about this, but it looks like you are basically saving the map name to the DB only to query it again a few lines later?

Content.Server/GameTicking/GameTicker.RoundFlow.cs
```diff
@@ -661,6 +663,7 @@ public void RestartRound()
             IncrementRoundNumber();
             SendRoundStartingDiscordMessage();
 
+
```
dffdff2423, line 666: Random whitespace in wizden code
```suggestion
```

Content.Server/Maps/GameMapManager.cs
```diff
@@ -190,12 +197,17 @@ public bool CheckMapExists(string gameMap)
 
     private bool IsMapEligible(GameMapPrototype map)
     {
-        return map.MaxPlayers >= _playerManager.PlayerCount &&
-               map.MinPlayers <= _playerManager.PlayerCount &&
-               map.Conditions.All(x => x.Check(map)) &&
+        return //map.MaxPlayers >= _playerManager.PlayerCount &&
+               //map.MinPlayers <= _playerManager.PlayerCount &&
+               //map.Conditions.All(x => x.Check(map)) &&
```
dffdff2423, line 202: I think we still want these conditions?

Content.Server.Database/CDModel.cs
```diff
@@ -10,6 +11,20 @@ namespace Content.Server.Database;
...
+    /// <summary>
+    /// Stores Advanced Round Information (shitcode probably idk)
+    /// </summary>
+    public sealed class AdvancedRound
+    {
+        [Key] public int Id { get; set; }
+        [ForeignKey("Server")] public int ServerId { get; set; }
+        [ForeignKey("Round")] public int RoundId { get; set; }
+        public string Map { get; set; } = null!;
+        public List<Player> Players { get; set; } = default!;
```
dffdff2423, line 23: Do we need this? I think the admin logs save this already?

dffdff2423 (changes requested):
Quite good for your first DB PR (mine certainly was worse). I just found a few things.

Additionally, the test fails are related to your changes.

Content.Server.Database/CDModel.cs
Aexxie: You're right, `Players` was left there by mistake while I was figuring out what I was doing.

Content.Server/Maps/GameMapManager.cs
Aexxie: Also a mistake, was using this during testing.

Content.Server/Database/ServerDbBase.cs
```diff
@@ -872,6 +874,51 @@ INSERT INTO player_round (players_id, rounds_id) VALUES ({players[player]}, {id}
...
+                .TakeLast(cacheDepth)
+                .Select(m => m.Map)
+                .ToArray();
+
+            foreach (var map in maps)
+            {
+                mapCache.Enqueue(map);
+            }
+
+            return maps;
```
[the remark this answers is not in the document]
Aexxie: Fuck, deleted my comment.
No need. `TakeLast` does exactly what it sounds like. The table is sorted in order of the default ID, which can only ever increase.

Content.Server/Database/ServerDbBase.cs
```diff
@@ -872,6 +874,51 @@ INSERT INTO player_round (players_id, rounds_id) VALUES ({players[player]}, {id}
...
+        public async Task<string[]> RetrieveMapQueue(Queue<string> mapCache, int cacheDepth)
+        {
+            await using var db = await GetDb();
+
+            var maps = db.DbContext.AdvancedRound
+                .Where(m => m.Map != "")
+                .ToList()
+                .TakeLast(cacheDepth)
+                .Select(m => m.Map)
+                .ToArray();
```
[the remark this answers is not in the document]
Aexxie: Fuck, deleted my comment.
No need. TakeLast does exactly what it sounds like. The table is sorted in order of the default ID, which can only ever increase.

Content.Server/Database/ServerDbBase.cs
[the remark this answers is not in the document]
Aexxie: There's no point in returning. It's a leftover from a previous approach I was taking.

Content.Server/_CD/GameTicking/GameTicker.AdvancedRound.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using Content.Shared._CD.CCVars;
+
+namespace Content.Server.GameTicking;
```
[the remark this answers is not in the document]
Aexxie: Sounds good. Wasn't sure. I was told to put everything like this in our namespace whenever possible.

Content.Server/_CD/GameTicking/GameTicker.AdvancedRound.cs
Aexxie: Unsure what you mean for the first part.

UpdateMapQueue will retrieve the last few maps from the db and return the query. This is to account for there being no maps in the query upon server start. MapCache will always have one element removed, and one element added, keeping it at a net change of 0. The initial condition will never be met as long as the queue has the correct amount of maps in it.

Content.Server.Database/CDModel.cs
```diff
@@ -10,6 +11,20 @@ namespace Content.Server.Database;
 
 public static class CDModel
 {
+    /// <summary>
+    /// Stores Advanced Round Information (shitcode probably idk)
+    /// </summary>
+    public sealed class AdvancedRound
```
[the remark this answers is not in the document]
dffdff2423: All CD tables should have a CD prefix to prevent conflicts. If you are making an extension onto a wizden table the convention is CDWizdenTable

Content.Server/GameTicking/GameTicker.RoundFlow.cs
```diff
@@ -147,6 +147,8 @@ private void LoadMaps()
                 if (i == 0)
                     DefaultMap = mapId;
             }
+
+            IncrementAdvancedRoundNumber(); // CD Addition. Caches current map and cycles queue, as well as adding an entry useful for statistics.
```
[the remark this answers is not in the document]
dffdff2423: Does it skip the automatically picked maps from those rounds in the vote?

Content.Server/Database/ServerDbBase.cs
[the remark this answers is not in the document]
dffdff2423: I think returning the maps here is a bit cleaner than taking the cache as the arg but I won't die on this hill.

Content.Server.Database/CDModel.cs
dffdff2423: You need to regenerate the migrations every time you change the model.

dffdff2423:
Hi, this has been sitting dormant for a while. Do you intend to finish it?

PR #593, opened 2025-04-09
Re-enable Space Dragons

Nairodian:
## About the PR
Re-enables the Space Dragon event at weight 1 because it was fixed Upstream to not delete bodies on butcher.

Also makes it a free agent and disables the 5 minute timer that deletes the dragon when no rifts are active. (Technically a buff.)

_What's the worst that could happen..?_

**Media**
image

## Requirements

**Changelog**
Space Dragons can now very rarely appear as a ghost role again and attempt to either conquer or befriend the station.

### DIFF
--- a/Content.Server/Dragon/DragonSystem.cs
+++ b/Content.Server/Dragon/DragonSystem.cs
@@ -97,12 +97,12 @@ public sealed partial class DragonSystem : EntitySystem
             if (!_mobState.IsDead(uid))
                 comp.RiftAccumulator += frameTime;
 
-            // Delete it, naughty dragon!
-            if (comp.RiftAccumulator >= comp.RiftMaxAccumulator)
-            {
-                Roar(uid, comp);
-                QueueDel(uid);
-            }
+            // Delete it, naughty dragon! #CD disable no rift deleting
+            // if (comp.RiftAccumulator >= comp.RiftMaxAccumulator)
+            // {
+            //     Roar(uid, comp);
+            //     QueueDel(uid);
+            // }
         }
     }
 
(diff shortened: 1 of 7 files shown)


dffdff2423 (approved):
Might be good to make it such that they don't need the portals to survive so it is *possible* to play a non-murderous dragon. However, I think this is fine as is.

Nairodian:
> Might be good to make it such that they don't need the portals to survive so it is _possible_ to play a non-murderous dragon. However, I think this is fine as is.

This is an interesting thought. In my mind you should be trying to progress your most reasonable goal as a ghost role. A spider would want to expand it's nest and stay fed. The slimes have been described as being irritated by everything around them, as such, always hostile.

So that opens the question, what exactly *are* reasonable goals for Space Dragons? The one that fits most with the game is that they're at the station to expand their territory. But because they're not forced onto the station like spiders or slimes, they have a lot of leeway here, and could have gone to the station for any means.

dffdff2423:
> So that opens the question, what exactly _are_ reasonable goals for Space Dragons? The one that fits most with the game is that they're at the station to expand their territory. But because they're not forced onto the station like spiders or slimes, they have a lot of leeway here, and could have gone to the station for any reason.

The main thing that I was thinking of when I said that was when dragons were last enabled, Aug played a greedy dragon that made a ton of demands of the station. A dragon is a scary enough threat alone that the station might seriously consider whatever demands that they have.

Nairodian:
> Might be good to make it such that they don't need the portals to survive so it is _possible_ to play a non-murderous dragon. However, I think this is fine as is.

Done.

dffdff2423 (approved):

LankLTE:
I think this could use some indication that space dragons are not *nice* beings even if they're technically free agents, if that makes sense, but that's tough to define and we could probably leave it as-is.

PR #599, opened 2025-04-14
Remove Intern Roles, Rework Them into Loadouts

Nairodian:
## About the PR
Removes intern roles from the game, intern PDAs can now be selected by the base department role it belongs to, adds +1 base role to every map.

(Bulldozes over some old code regarding the roboticist PDA, I'll clean it up in a later PR.)

This is *probably* a fine idea and execution.

## Media
image

## Requirements

**Changelog**
Intern roles have been removed, however instead they have been reworked into job titles you can pick via loadouts.


LankLTE:
Did we ever get to a definite decision on how we wanted to handle this? I think this is a fine idea tbh though some people had concerns that it was too hidden

LankLTE:
Waiting until after the upstream merge since that touches map stuff

dffdff2423:
> Did we ever get to a definite decision on how we wanted to handle this? I think this is a fine idea tbh though some people had concerns that it was too hidden

I think I have said I was not a fan of this in the past due to the possibility for new player confusion. We have also been getting a lot of new to ss14 players recently.

Nairodian:
> I think I have said I was not a fan of this in the past due to the possibility for new player confusion. We have also been getting a lot of new to ss14 players recently.

Alright there's no true answer here. We strongly want to merge the roles, so that's already a must. But the way I see it is that you risk less by using the regular as the default. Obviously this mean that new players are less likely to show up as interns, but that's less of an issue of a regular player showing up as an intern because they forgot to set their loadout. Not only will it affect the majority of people less, it's also just less damaging in the grand scheme of things.

The seniors should probably still be checking with regular staff to ensure everyone's well versed, and most new players will ask for help. In the worst case if we want an alternate solution, seniors/command staff could get a HUD that marks new players, either to their role or to the game totally. It's a little immersion breaking but it theoretically wouldn't be seen often.

PR #605, opened 2025-04-24
admin records console can filter by record entry type again

lunarcomets:
## About the PR

fixes admin record console not showing the proper record entries when you select an entry filter (for filtering security, employment, or medical records)

## Requirements

- [X] I have read and I am following the Cosmatic Drift [PR Guidelines](https://github.com/cosmatic-drift-14/cosmatic-drift/blob/master/CONTRIBUTING.md) (note that they may differ than what you find on other SS14 forks).
- [X] I have approval from a maintainer if this PR adds a feature OR this PR does not add a feature OR you are alright with this PR being closed at a maintainer's discression.
- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

ADMIN: the admin records console now actually lets you filter what records you want to look at again via record type

PR #606, opened 2025-04-24
admin records

lunarcomets:
## About the PR

adds admin records for players to add to their characters that are only visible to game admins!

meant to allow for players to put information about their characters into this section that would only be seen by game admins and the character's player OOC, and by respective corporations, organizations, and individuals IC without having to relay all that information through ahelps or DMs, and instead by pointing to a particular record if needed
requires #605 to be merged to actually allow admins to see confidential records (for now)(done)

## Media

image
image
image

## Requirements

**Changelog**

Admin records have been added for entries that are meant to be view by the player and game admins only!

for the admins: the admin record console has a new filter that allows admins to see a player's admin records, visible only to the GM team and that player

### DIFF
--- a/Content.Server.Database/Migrations/Postgres/20240617225327_CDRecordJsonToTable.cs
+++ b/Content.Server.Database/Migrations/Postgres/20240617225327_CDRecordJsonToTable.cs
@@ -79,9 +79,9 @@ namespace Content.Server.Database.Migrations.Postgres
                 INSERT INTO cd_character_record_entries (title, involved, description, type, cdprofile_id)
                     SELECT
                         jsonb_array_elements.value ->> 'Title', jsonb_array_elements.value ->> 'Involved', jsonb_array_elements.value ->> 'Description',
-                        {(int)CDModel.DbRecordEntryType.Confidential}, cdprofile_id
+                        {(int)CDModel.DbRecordEntryType.Admin}, cdprofile_id
                     FROM
-                        cdprofile, jsonb_array_elements(character_records -> 'ConfidentialEntries')
+                        cdprofile, jsonb_array_elements(character_records -> 'AdminEntries')
                 """);
         }
 
--- a/Content.Server.Database/Migrations/Postgres/20240617225327_CDRecordJsonToTable.cs
+++ b/Content.Server.Database/Migrations/Postgres/20240617225327_CDRecordJsonToTable.cs
@@ -74,6 +74,15 @@ namespace Content.Server.Database.Migrations.Postgres
                     FROM
                         cdprofile, jsonb_array_elements(character_records -> 'EmploymentEntries')
                 """);
+
+            migrationBuilder.Sql($"""
+                INSERT INTO cd_character_record_entries (title, involved, description, type, cdprofile_id)
+                    SELECT
+                        jsonb_array_elements.value ->> 'Title', jsonb_array_elements.value ->> 'Involved', jsonb_array_elements.value ->> 'Description',
+                        {(int)CDModel.DbRecordEntryType.Confidential}, cdprofile_id
+                    FROM
+                        cdprofile, jsonb_array_elements(character_records -> 'ConfidentialEntries')
+                """);
         }
 
         /// <inheritdoc />
--- a/Content.Server.Database/Migrations/Sqlite/20240617225311_CDRecordJsonToTable.cs
+++ b/Content.Server.Database/Migrations/Sqlite/20240617225311_CDRecordJsonToTable.cs
@@ -76,9 +76,9 @@ namespace Content.Server.Database.Migrations.Sqlite
                 INSERT INTO cd_character_record_entries (title, involved, description, type, cdprofile_id)
                     SELECT
                         json_each.value ->> '$.Title', json_each.value ->> '$.Involved', json_each.value ->> '$.Description',
-                        {(int)CDModel.DbRecordEntryType.Confidential}, cdprofile_id
+                        {(int)CDModel.DbRecordEntryType.Admin}, cdprofile_id
                     FROM
-                        cdprofile, json_each(character_records, '$.ConfidentialEntries')
+                        cdprofile, json_each(character_records, '$.AdminEntries')
                 """);
         }
 
--- a/Content.Server.Database/Migrations/Sqlite/20240617225311_CDRecordJsonToTable.cs
+++ b/Content.Server.Database/Migrations/Sqlite/20240617225311_CDRecordJsonToTable.cs
@@ -71,6 +71,15 @@ namespace Content.Server.Database.Migrations.Sqlite
                     FROM
                         cdprofile, json_each(character_records, '$.EmploymentEntries')
                 """);
+
+            migrationBuilder.Sql($"""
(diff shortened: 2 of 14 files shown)


dffdff2423:
I don't think "Confidential" makes it exactly clear it is intended for admins. Maybe, something like "Admin Info" or "OOC" would be better?

dffdff2423:
You included the bugfix with this branch. There is a decent chance it will merge conflict once the other PR is merged.

lunarcomets:
> I don't think "Confidential" makes it exactly clear it is intended for admins. Maybe, something like "Admin Info" or "OOC" would be better?

the intention with 'Confidential' is to try to keep it IC in some sense and to still have it be able to be referred to IC

would 'Confidential (Admin Info)' work, though?

Content.Shared/_CD/Records/PlayerProvidedCharacterRecords.cs
```diff
@@ -100,7 +102,10 @@ public PlayerProvidedCharacterRecords(
         string identifyingFeatures,
         string allergies, string drugAllergies,
         string postmortemInstructions,
-        List<RecordEntry> medicalEntries, List<RecordEntry> securityEntries, List<RecordEntry> employmentEntries)
+        List<RecordEntry> medicalEntries,
```
dffdff2423, line 105: Can you replace this constructor with a primary constructor?

lunarcomets:
also i rebased this branch on top of the bugfix branch so i could actually test things; can i just omit the bugfix commits somehow?

dffdff2423 (changes requested):
Nothing majorly wrong on first read. I will give it another look later. Did you test to make sure that it worked properly with characters created before confidential records existed in the codebase?

dffdff2423:
> also i rebased this branch on top of the bugfix branch so i could actually test things; can i just omit the bugfix commits somehow?

uh.... you can if you like but it is probably easier to just wait till we merge your bugfix and fix whatever happens as a result of it.

dffdff2423:
> the intention with 'Confidential' is to try to keep it IC in some sense and to still have it be able to be referred to IC

I am not sure that making them IC at all is a good idea. People can include stuff that is intended to be viewed by CC in them but it should not just be limited to IC stuff. Can you also update the guidebook to include some info about them once we hammer it out? 

> would 'Confidential (Admin Info)' work, though?
That seems a bit wordy... I would also like something that get's the point across that they should include character blackmail (and other things antags can take advantage of) there.

lunarcomets:
> That seems a bit wordy... I would also like something that get's the point across that they should include character blackmail (and other things antags can take advantage of) there.

i'm not sure, if you can could you run it by the GMs for me for thoughts? i'm pulling a blank for what to have it called if Confidential won't work, since Admin Info or OOC doesn't feel right either imo

Content.Shared/_CD/Records/PlayerProvidedCharacterRecords.cs
lunarcomets: the entire `PlayerProvidedCharacterRecords` constructor?

lunarcomets:
> Nothing majorly wrong on first read. I will give it another look later. Did you test to make sure that it worked properly with characters created before confidential records existed in the codebase?

also just tested: works fine with already existing characters, but there's an error when importing an old CD character profile without confidential records to the server - looks like it is because the confidential records are null and `EnsureValidEntries` doesn't take into account null fields

dffdff2423:
> also just tested: works fine with already existing characters, but there's an error when importing an old CD character profile without confidential records to the server - looks like it is because the confidential records are null and `EnsureValidEntries` doesn't take into account null fields

If you are talking about YAML file imports that is fine as long as it does not crash

lunarcomets:
then it's fine, just need to fix a primary constructor error now when i get home then

Content.Shared/_CD/Records/PlayerProvidedCharacterRecords.cs
lunarcomets: because of [this](https://stackoverflow.com/questions/65271665/why-is-an-explicit-this-constructor-initializer-required-in-records-with-a-pri), i'll just mark this as resolved

Ecramox:
This as an option in the character creator, is not IC info. There is no particular need to name it IC, and if anyone wants to refer to these records IC, they can choose whatever they want to call them. I would like to just name it Admin Records so it's clear, leaves no room for misinterpretation. 

It also means people won't feel like they'll have to write them IC if they don't want to, since this will be viewed solely from an admin perspective and not an IC perspective, which depending on the person, will lead to far clearer notes on what the admins should know about a character. It also means people will know that they can write down stuff that isn't necessarily known by Nanotrasen without violating their lore in any way.

Being as clear as possible on what is IC and OOC information is always ideal, since it leads to much clearer communication overall.

dffdff2423:
The general consensus on discord was that they should just be called Admin records.

lunarcomets:
added a guidebook entry and changed this into just 'admin records', should only need approval of the guidebook entry now

Content.Server.Database/Migrations/Postgres/20240617225327_CDRecordJsonToTable.cs
```diff
@@ -74,6 +74,15 @@ INSERT INTO cd_character_record_entries (title, involved, description, type, cdp
...
                 """);
+
+            migrationBuilder.Sql($"""
+                INSERT INTO cd_character_record_entries (title, involved, description, type, cdprofile_id)
+                    SELECT
+                        jsonb_array_elements.value ->> 'Title', jsonb_array_elements.value ->> 'Involved', jsonb_array_elements.value ->> 'Description',
+                        {(int)CDModel.DbRecordEntryType.Admin}, cdprofile_id
+                    FROM
+                        cdprofile, jsonb_array_elements(character_records -> 'AdminEntries')
+                """);
```
dffdff2423, line 85: These migrations are not needed and I would prefer they be excluded if you did not test them. They would only be used in the event a fork cheery picks this onto a version of records when entries were still stored in the JSON and then merges the pr that turned them into a table which is just very unlikely.

Content.Server.Database/Migrations/Sqlite/20240617225311_CDRecordJsonToTable.cs
```diff
@@ -71,6 +71,15 @@ INSERT INTO cd_character_record_entries (title, involved, description, type, cdp
...
                 """);
+
+            migrationBuilder.Sql($"""
+                INSERT INTO cd_character_record_entries (title, involved, description, type, cdprofile_id)
+                    SELECT
+                        json_each.value ->> '$.Title', json_each.value ->> '$.Involved', json_each.value ->> '$.Description',
+                        {(int)CDModel.DbRecordEntryType.Admin}, cdprofile_id
+                    FROM
+                        cdprofile, json_each(character_records, '$.AdminEntries')
+                """);
```
dffdff2423, line 82: same as above

dffdff2423 (changes requested):
Guidebook looks good, I would prefer this be merged after I fix the current issues with records (I will start working on that after work today).

dffdff2423 (approved):
I did not get to fixing them, I thought I had time I did not have. Feel free to merge whenever Lank

PR #595, opened 2025-04-10
Porting all Impstation markings. None are made by me.

Ecramox:
## About the PR
A few hundred markings from Impstation! Upstream markings are completely barren, and it hurts me so. None of these are made by me. Permission was given by all artists for this port.
Credit for the markings goes to (Names as listed in the meta.jsons):
Carousel
Sha-Seng
Crekulon
Honeyed_Lemons_
Doop
Skeleskitty
Snavian
pinkbat5
starlighthowls
rainyatrium
LandsharkRAWR

## Media
There's uh, a lot. I'll have some pictures of the menu and what a few characters look like with these markings, but I'd recommend putting this PR on a private branch and playing around with it to see if there's anything you'd change.
(Also, I don't know off the top of my head which markings are upstream and which are Imp, so a few upstream markings may be in these screenshots. Sorry!)
image
image
image
image
image

## Requirements

- [X] I have read and I am following the Cosmatic Drift [PR Guidelines](https://github.com/cosmatic-drift-14/cosmatic-drift/blob/master/CONTRIBUTING.md) (note that they may differ than what you find on other SS14 forks).
- [X] I have approval from a maintainer if this PR adds a feature OR this PR does not add a feature OR you are alright with this PR being closed at a maintainer's discression.
- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

PR #637, opened 2025-05-17
Impstation Diona Markings

Ecramox:
## About the PR
From Impstation, ports the several new Diona markings. Also ports changes to upstream ones, being that they're now grayscaled (So you can color them better for color schemes other than realistic green plant) and adds a few new color channels, so you can get more funky with said coloring!

## Media
image
image

## Requirements

**Changelog**
Due to having grayscale markings, Diona are now more colorful than ever!
New Diona markings added!


Ecramox:
BTW, permission from all creators is in the comments of the original markings hell PR.
https://github.com/cosmatic-drift-14/cosmatic-drift/pull/595

Resources/Prototypes/Entities/Mobs/Customization/Markings/diona.yml
```diff
@@ -3,56 +3,40 @@
...
   markingCategory: Chest
   speciesRestriction: [Diona]
-  coloring:
-    default:
-      type:
-        !type:SimpleColoring
-          color: "#FFFFFF"
   sprites:
-  - sprite: Mobs/Customization/diona.rsi
-    state: thorns_body
```
dffdff2423, line 27: Correct me if I am wrong, but it looks like you are getting rid of wizden markings here. Is this needed?

Ecramox: The only difference between the two is that it's more colorable. This does not erase past possibilities, it only expands them. Every Diona that existed before can still exist with these changes, and more.

The only minor sticking point is that currently made Diona's might have their color changed, but that's easy enough to fix.

dffdff2423: I am fine with that, not sure if @LankLTE would be. Those should be marked as a CD change though.

LankLTE: Comment things out instead of deleting them.

LankLTE (changes requested):

Ecramox:
RSI failure isn't in a file I touched, so I'm going to assume it isn't my fault.

PR #689, opened 2025-07-06
PAIs can talk on binary

dffdff2423:
## About the PR

Makes PAIs able to talk on the binary channel with `:b`. It makes sense and adds additional utility (eg. a PAI could report salvage has died).

## Requirements
- [ ] I have read and I am following the Cosmatic Drift [PR Guidelines](https://github.com/cosmatic-drift-14/cosmatic-drift/blob/master/CONTRIBUTING.md) (note that they may differ than what you find on other SS14 forks).
- [ ] I have approval from a maintainer if this PR adds a feature **or** this PR does not add a feature **or** I am alright with this PR being closed at a maintainer's discretion.
- [ ] I have added screenshots/videos to this PR showcasing its changes ingame **or** this PR does not require an ingame showcase.

**Changelog**

PAIs can talk on binary with :b


dffdff2423:
> The reason this wasn't done upstream is that it becomes "optimal" to have a pAI at all times, because if anything happens to the player the pAI can immediately report it. That's mostly the case for antags- but I would make the argument that salvage wanting to have a pAI at all times so that they can report if they die is equally as bad

I don't think that is a bad thing. pAIs are rarely used on CD and almost every time salvage dies CC is faxed for coords if the station does not know where they are.

PR #716, opened 2025-07-28
Reduce the volume of the tech disk console

dffdff2423:
## About the PR

Lowered the volume of the tech disk console because Nairod asked me to.

**Changelog**

Gambling (the teck disk dispenser) no longer destroys your ear drums.


LankLTE:
1984

Issue #725, opened 2025-08-06
Allergies database migration breaks on Postgres


TsjipTsjip:
## Description

The database migration in [this file](https://github.com/cosmatic-drift-14/cosmatic-drift/blob/master/Content.Server.Database/Migrations/Postgres/20250609134446_CDAllergies.cs) which was introduced by #649 does not match its SQLite counterpart introduced by the same PR.

The current Postgres migration causes a server running Postgres to crash.

**Reproduction**

Run the server with a Postgres database attached.

**Additional context**

I am not certain how the migration was generated, but the Up method attempts to remove an index by name `IX_cd_character_allergies_allergen` on the `cd_character_allergies` table, where neither exist in the database state when this action is attempted. The migration does not generate this table either.

This is a contrast to the SQLite counterpart of this migration which creates the table first as expected.

PR #649, opened 2025-06-03
Add Allergies

ashkitten:
## About the PR

Add customizable allergies: reagents that when ingested, imbibed, inhaled, injected, or contacted produce an adverse reaction in the form of histamine production. This solves a longstanding issue of allergies being present in roleplay but not represented in gameplay mechanics. Now, instead of having to tell people that you're having trouble breathing, you can simply collapse on the ground gasping for air!

## Media

image
2025-06-03 13-50-33.webm

## Requirements

- [x] I have read and I am following the Cosmatic Drift [PR Guidelines](https://github.com/cosmatic-drift-14/cosmatic-drift/blob/master/CONTRIBUTING.md) (note that they may differ than what you find on other SS14 forks).
- [x] I have approval from a maintainer if this PR adds a feature OR this PR does not add a feature OR you are alright with this PR being closed at a maintainer's discression.
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

Added allergic reactions. Now perish!

PR #727, opened 2025-08-06
Fix allergies migration

dffdff2423:
## About the PR

Fix the postgres migration for #649. 

Copy the sqlite migration to the postgres migration. It seemed to work
correctly on my test db but as always make sure to take a backup before you
apply DB migrations

I would like to take this moment to rant about EF core. The migration generator
shits itself if you look at it in the wrong way and is very not deterministic.
Git restoring the files then running the migration generator again can produce
different results. On top of that, EF core is a massive pain to debug because
of all the ""Nice abstractions"" or whatever and the documentation is
gradle-esque documentation theater.

Fixes: #725


TsjipTsjip (approved):
This PR resolves the linked issue and allows servers using Postgres to start again.

PR #735, opened 2025-08-14
Catatonic and Soulless description changes

SpaceLizard24:
## About the PR
Changed the descriptions for soulless and catatonic players to explain that they will recover, just not on the station. This is so that ~~i can get contrib role :3~~ players don't get confused about how to deal with catatonic and soulless players. 

Changes are below. CC having the magic, catatonic revival medicine or treatment seems fine enough. Cryopods probably bring you to CC or could be brought to CC, and obviously the evac shuttle brings you to CC. 

Encouraging players to revive soulless people seems like the right thing to do. It's probably easier to wake up an alive but catatonic person, rather than cloning someone from scratch, because the players didn't know they needed to be revived.

## Media
image
image

## Requirements
- [X] I have approval from a maintainer if this PR adds a feature **or** this PR does not add a feature **or** I am alright with this PR being closed at a maintainer's discretion.
- [X] I have added screenshots/videos to this PR showcasing its changes ingame **or** this PR does not require an ingame showcase.

**Changelog**
Changed the description of catatonic and soulless players.

### DIFF
--- a/Resources/Locale/en-US/mind/components/mind-component.ftl
+++ b/Resources/Locale/en-US/mind/components/mind-component.ftl
@@ -4,8 +4,8 @@ comp-mind-ghosting-prevented = You are not able to ghost right now.
 
 ## Messages displayed when a body is examined and in a certain state
 
-comp-mind-examined-catatonic = { CAPITALIZE(SUBJECT($ent)) } { CONJUGATE-BE($ent) } totally catatonic. The stresses of life in deep-space must have been too much for { OBJECT($ent) }. Any recovery is impossible.
+comp-mind-examined-catatonic = { CAPITALIZE(SUBJECT($ent)) } { CONJUGATE-BE($ent) } totally catatonic. { CAPITALIZE(SUBJECT($ent)) } will not recover without specalized treatment, available at Central Command. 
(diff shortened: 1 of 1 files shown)


LankLTE:
Feel the need to point out negatively interacting with SSD players is already against the rules, so I don't see why we need to change text to discourage that. 

This text specifically is also problematic as it implies they can be revived at roundend CC, which is something we're obviously not capable of doing.

Nairodian:
I agree with Lank that the text doesn't require changing, but there's no reason not to either. If you're going to change it, do, "They may take a few days to recover" or something along those lines so it's clear that it's unlikely to happen in round, but still encourages them to keep them.

SpaceLizard24:
updated it: changed it from saying "cc has medicine" to "yeah itll take a little bit, off station
"

Nairodian (changes requested):
Nitpick but change it from will to may because occasionally admins will control catatonic entities, or players will take ghost roles then ahelp to be sent back to their original bodies after.

SpaceLizard24:
may

Nairodian (approved):

dffdff2423 (approved):

PR #748, opened 2025-09-14
Replace Ion Storms notifying synths with Synth Storm event

LankLTE:
## About the PR

Replaced Ion Storms having a chance to notify synths with a new "synth storm" event, which runs seperately and will only notify one synth at a time. The notification may want / need to be changed but I'll leave that up to Nairod. Also added a sound to make it more obvious when it happens.
Ion Storms are incredibly common to spice up borg gameplay, but that also meant synths would frequently be going haywire which made them look unreliable from an IC standpoint and could cause a lot of headache for robotocists and the like for frequently having to put up with them. This should help mitigate that while also preventing synth players from metagaming that borgs have been ion stormed.


dffdff2423:
An idea which was mentioned a while ago was to make all of the synths in a ~10 tile radius also stormed.

dffdff2423 (approved):

PR #527, opened 2024-12-30
Snake Tails

KittenColony:
## About the PR
Ports the 5 Snake Tails from Bubberstation for Lizards to use.

## Why / Balance
Character expression! More tail options for lizards that aren't the regular style

## Technical details
Adds 18 marking sprites with attached YAML and FTL files

## Media
image
image
image
image
image
image
image

## Requirements

- [x] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**
🆑 

Add: Adds Five new lizard tail options, Snake tails. Plain, Stripes, Stripes Alt, Under and Dual

PR #755, opened 2025-09-27
Disable lizard hardsuit bodies and hide hair

dffdff2423:
## About the PR

Disables lizard hardsuit body sprites.

While the sprites created in https://github.com/space-wizards/space-station-14/pull/35842 look great, they kind of break continuity with lizards that have large tails (from wizden), or the tails added in #527 (which are a fairly popular option).

Sooner or later, I wager somebody will code a proper hardsuit marking system which would fix this issue.

Disabled using a few wads of shitcode rather than editing the meta files to prevent merge conflicts.

I also went and made Moth and Lizard hair get hidden upon using a hardsuit helmet. Probably should have been a separate PR but too late now.

**Changelog**

- Temporarily (possibly permanently) disable the lizard hardsuit sprites until a more robust hardsuit marking sprite system is created.
- Moth and Lizard hair is now properly hidden when you put on a helmet. This is the same behavior as humans *et al*.

### DIFF
--- a/Content.Client/Clothing/ClientClothingSystem.cs
+++ b/Content.Client/Clothing/ClientClothingSystem.cs
@@ -165,11 +165,14 @@ public sealed class ClientClothingSystem : ClothingSystem
         if (clothing.EquippedState != null)
             state = $"{clothing.EquippedState}";
 
-        // species specific
-        if (speciesId != null && rsi.TryGetState($"{state}-{speciesId}", out _))
-            state = $"{state}-{speciesId}";
-        else if (!rsi.TryGetState(state, out _))
-            return false;
+        // CD: Shitcode W (hide lizard hardsuit sprites)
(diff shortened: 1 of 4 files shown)


TiniestShark:
Yeah apologies on the singular tail option, I genuinely tried to figure out how to get something separated from the suit that could be modified for other forks (I wanted upstream to have a consistent singular tail on the suits so liz suits would all blend together like human suits) but I couldn't figure it out.

TiniestShark:
> No worries. There is no feature in the game right now that would let you do that as far as I am aware. It would require writing quite a bit of c# to add support for different tails and I am not sure wizden would even accept a PR which did that (years ago they explicitly said they did not want a system like this but I have no clue what the current stance is).

Something that could swap parts for different species (Arachnid arms, moth wings, liz tails, etc) would be accepted, but individual sprites for each marking likely wouldn't.

dffdff2423:
image

PR #773, opened 2025-10-19
Revert Ambrosia Losing Desox

Nairodian:
## About the PR
The only reasons listed in https://github.com/space-wizards/space-station-14/pull/40638 do not apply to CD due to the different handling of illegal substances in our SOP. Alongside the couple complaints that locking the botanical production of Desox behind mutations was significantly problematic, I can see valid reason to revert the change back to Ambrosia.

If a valid balance, design, or flow concern can be made as to why Glasstle is more preferable to have Desox than Ambrosia, sure, but for now I think we may as well revert it.

## Media
image

## Requirements
- [X] I have approval from a maintainer if this PR adds a feature **or** this PR does not add a feature **or** I am alright with this PR being closed at a maintainer's discretion.
- [X] I have added screenshots/videos to this PR showcasing its changes ingame **or** this PR does not require an ingame showcase.

**Changelog**
Reverted the change that moved Desoxyephedrine from Ambrosia to Glasstle.

### DIFF
--- a/Resources/Prototypes/Hydroponics/seeds.yml
+++ b/Resources/Prototypes/Hydroponics/seeds.yml
@@ -1294,7 +1294,7 @@
       Min: 1
       Max: 5
       PotencyDivisor: 20
-    Desoxyephedrine: #CD: Moved Desox back to ambrosia
+    Desoxyephedrine: #CD: Moved Desox back to ambrosia from glasstle.
       Min: 1
       Max: 10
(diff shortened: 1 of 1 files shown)


dffdff2423:
Narcolepsy is something which players opt into so I am not too concerned about that. I saw the suggestion somewhere of allowing people to take a few pills as a loadout and I think that is a better solution.

PR #817, opened 2026-01-08
Revert Pipe Stacking Removal

LankLTE:
This was reverted on our old master.


dffdff2423:
Maybe we should try keeping the pipe stacking ban now that manifolds exist?

dffdff2423:
I don't play atmos so I will take your word for it

PR #834, opened 2026-01-17
add pickable brains for cyborgs

lunarcomets:
## About the PR

adds the ability for players to pick a borg's brain type via the borgs loadout menu from:
filled mmi (standard/human brain)
positronic brain
circuit-based brain 

also adds a circuit-based brain for customization purposes

codewise, the new `CdPlayerSpawnBeforeMindEvent` is now raised right before any entity is given their mind if being spawned for a player to take over, primarily useful for when this system expands to AIs (and possiblty normal players as well to change what organs they hold?) mostly done for this pr though so we don't get a bunch of light flickers on round start that otherwise would happen with other events

## Media

https://github.com/user-attachments/assets/f53b75c5-15e0-462f-9f62-508e51e23ca4

## Requirements

**Changelog**

Cyborg brains are now customizable via loadouts. Comes with a new brain type!

### DIFF
--- a/Content.Server/_CD/Silicons/SiliconBrainLoadoutSystem.cs
+++ b/Content.Server/_CD/Silicons/SiliconBrainLoadoutSystem.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics;
 using Content.Server.GameTicking;
 using Content.Server.Players;
 using Content.Server.Preferences.Managers;
@@ -6,6 +7,7 @@ using Content.Shared.GameTicking;
 using Content.Shared.Mind;
 using Content.Shared.Preferences;
 using Content.Shared.Preferences.Loadouts;
+using Content.Shared.Roles;
 using Content.Shared.Silicons.Borgs.Components;
 using Content.Shared.Station;
 using Robust.Shared.Containers;
@@ -25,6 +27,8 @@ public sealed class SiliconBrainLoadoutSystem : EntitySystem
     [Dependency] private readonly GameTicker _gameTicker = default!;
     [Dependency] private readonly IPrototypeManager _proto = default!;
 
+    private ProtoId<LoadoutGroupPrototype> CyborgBrainLoadoutPrototype => "CyborgBrain";
+
     public override void Initialize()
     {
         base.Initialize();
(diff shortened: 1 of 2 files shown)


Content.Server/_CD/Silicons/SiliconBrainLoadoutSystem.cs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    private void OnBorgPlayerSpawnComplete(Entity<BorgChassisComponent> ent, ref CdPlayerSpawnBeforeMindEvent args)
+    {
+        // surely there's a better way of doing this
+        var jobLoadoutId = LoadoutSystem.GetJobPrototype(args.JobId);
+        var loadoutProto = args
+            .Character
+            .GetLoadoutOrDefault(jobLoadoutId, args.Player, args.Character.Species, EntityManager, _proto)
+            .SelectedLoadouts["CyborgBrain"][0].Prototype;
```
dffdff2423: If you can, move the string to a component or a ProtoId<T> constant so it is validated. Not 100% sure that works with loadouts.

Also, why can we get away with just taking the 0th element here rather than looping over? If this is an expected precondition it should be asserted.

Content.Server/_CD/Silicons/SiliconBrainLoadoutSystem.cs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/// <summary>
+/// Event that raises right before the mind is added to the player entity while spawning on station.
+/// </summary>
+public sealed class CdPlayerSpawnBeforeMindEvent(ICommonSession player, HumanoidCharacterProfile character, string jobId) : EntityEventArgs
+{
+    public ICommonSession Player { get; } = player;
+    public HumanoidCharacterProfile Character { get; } = character;
+    public string JobId { get; } = jobId;
```
dffdff2423: Would this work as a `ProtoId<JobPrototype>`?

You also might be able to make this a bit nicer by using a record instead of a class but it is not a big deal.

dffdff2423 (changes requested):
Looks good. Just a few minor nit beyond the failing tests.

Content.Server/_CD/Silicons/SiliconBrainLoadoutSystem.cs
lunarcomets: it's an already expected precondition yeah, defined (and hopefully limited limited by) the min and max of the CyborgBrain loadoutGroup proto, if it has more than 1 something has gone terribly wrong

PR #892, opened 2026-04-26
Fix Candy Cigs

Nairodian:
## About the PR
Makes it so the sugar rush candy cigarettes don't immediately extinguish and error upon trying to light them.
Does light clean-up work on the prototype file as well.

## Media
I don't know what the actual desired direction for these is, I just don't want them to be showing up as errors.
image

## Requirements

**Changelog**
Sugar Rush candy cigarettes will no longer error upon lighting them.

### DIFF
--- a/Resources/Prototypes/_CD/Entities/Objects/Consumable/Smokeables/Cigarettes/cigarette.yml
+++ b/Resources/Prototypes/_CD/Entities/Objects/Consumable/Smokeables/Cigarettes/cigarette.yml
@@ -16,28 +16,28 @@
             Quantity: 5
 
 - type: entity
-  parent: [ Cigarette, FoodBase ]
+  parent: Cigarette
   id: CigaretteCandy
   name: cigarette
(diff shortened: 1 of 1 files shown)


lunarcomets (approved):
i think the purpose of the candy cigarettes are just to be bits of candy that look like cigarettes that you can put in your mouth, but can't actually 'smoke', hence the foodbase parent

to the point of fixing the error this is fine, but if you'd like you could make it so this works as a wearable piece of food that fits in the mouth, while getting rid of all the cigarette functionality so it can't be lit

otherwise, looks good

LankLTE:
I thought the point was just to give you the capability to aura farm without actually having to have your character be a smoker, ngl. To that extent I'm happy with this.

KittenColony:
Its so cigarette companies can normalize smoking thus increasing revenue.

And so people can IC pretend to smoke
Or just enjoy candy bc this is candy

PR #905, opened 2026-06-21
Merge master 210626

dffdff2423:
Opted for commit bf09298 since wizden has not released a new stable branch and that was the latest commit that passes tests.

**NOTE FOR CONTRIBUTORS- DO A PING AFTER MERGE**

We changed from PJB-den to Wizden RT in this commit. Your git repo may have some weirdness if you just try to do a git pull. You can fix it by removing `RobustToolbox` and `.git/modules/RobustToolbox` then editing `.git/config` to point to `space-wizards-federation` instead of `space-wizards`. Finally rerun `RUN_THIS.py` or init submodules manually. You may need to enable "show hidden files" in your OS settings to see the `.git` directory in the root of the CD repo.

You can also just nuke your local git repo and re-clone CD if you will not loose any local progress.


dffdff2423:
Looks fine to merge

PR #909, opened 2026-06-30
Marking Flares

Nairodian:
## About the PR
Adds a marking flare item that produces a small light for an hour before burning out.
-Comes in 4 different colors.
-Found in pouches, each containing 16 flares.
-2 pouches in the salvage vendor.
-Crates of 4 pouches purchasable for 900 spesos.
-Flares take up 1 inventory square, pouches take 2 meaning they can fit in your pocket.

I felt that there was a missing niche for long term light sources that you could toss around and stuff. The primary thing that comes to mind is salvagers being able to mark their path as they make it through things like the VGroid, so that they can find their way back to their ship without a mass scanner, or so that if they die, people can follow their path of flares to their corpse.

(Reorganized the tags file for CD to match Upstream while I was there anyways.)

## Media
image
image

## Requirements

**Changelog**
Marking flares have been added, which are small flares that last up to an hour before burning out. The salvage vendor starts with two pouches of them and more can be ordered through cargo.


dffdff2423 (approved):

lunarcomets:
the fact that those repeated slots are the only way to make the pouches work as you want them to right now makes me cry

good for merge though, i'll just add a TODO regarding the slots before i pull the lever

PR #917, opened 2026-07-20
allow AIs to freeform emote through holopads

lunarcomets:
## About the PR

what it says on the tin

this gives _only_ the AI the ability to emote through the holopads, holopads in use by anyone else act as normal

## Media

image

## Requirements

**Changelog**

AIs can now emote through holopads.

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.PrivateAPI.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.PrivateAPI.cs
@@ -207,6 +207,11 @@ public sealed partial class ChatSystem
                 _adminLogger.Add(LogType.Chat, LogImpact.Low, $"Emote from {source} as {name}: {action}");
             else
                 _adminLogger.Add(LogType.Chat, LogImpact.Low, $"Emote from {source}: {action}");
+
+        // CD - allow us to listen to any emotes
+        var ev = new CDEntityEmotedEvent(source, action);
+        RaiseLocalEvent(source, ref ev, true);
+        // end CD
     }
 
     // ReSharper disable once InconsistentNaming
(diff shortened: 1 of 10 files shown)


Nairodian:
Is it easy to make it transmit anyone's emotes? Because a lot of people already act like it does and it really should.

lunarcomets:
> Is it easy to make it transmit anyone's emotes? Because a lot of people already act like it does and it really should.

it would just mean setting a single field true for the holopad proto so yeah, pretty easy

i'm just not sure about doing it here because of potential issues with the holopad being more difficult to use in group calls

dffdff2423:
> i'm just not sure about doing it here because of potential issues with the holopad being more difficult to use in group calls

What exactly is your thoughts here?

lunarcomets:
> > i'm just not sure about doing it here because of potential issues with the holopad being more difficult to use in group calls
> 
> What exactly is your thoughts here?

whenever a group call happens over holopad, like a command meeting with CC for example, i tend to notice that a lot of people use emotes to interact with each other throughout the call a lot, since they can do it silently and without having the holopad pick up on that emote

having holopads be able to pick up those emotes with the way the system works right now would mean that everyone in range of the holopad would now no longer be able to talk or emote unless they don't mind spamming the holopad, which can make it really hard to follow for whoever's listening. preferably, holopads would be able to lock onto a person only through their voice or explicitly, and then relay their emotes based on if it's the right person, but that's not in the scope of this pr

AIs just don't have that problem since they always call alone, so they can emote without issue

Nairodian:
Bike-shedding, maybe you should just be able to toggle holopads between close-range listening (3x3 tile space) and long-range listening (whatever it is now), that way we'd have the best of both worlds.
(Also holopads should really give positive feedback, like "message transmitted!" when you successfully speak into them but that's a topic for another day.)

As it stands right now, it is probably better to keep emotes from non-AI users for the group call thing.

PR #921, opened 2026-08-16
Grit metabolisation rate decreased

Makar0v-fr:
## About the PR

This changes the metabolization rate from 0.5 to 0.2. The time it takes to produce it isn't worth it
with its current rate. Also it will make people useless of it since you can't realy make big batches of it.

## Requirements

- [x] I have read and am following the [Wizard's den Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I understand this PR may be closed if I did not seek prior approval for content additions.

**Changelog**

### DIFF
--- a/Resources/Prototypes/_CD/Reagents/narcotics.yml
+++ b/Resources/Prototypes/_CD/Reagents/narcotics.yml
@@ -422,7 +422,7 @@
   physicalDesc: reagent-physical-desc-volatile
   metabolisms:
     Bloodstream:
-      metabolismRate: 0.5
+      metabolismRate: 0.2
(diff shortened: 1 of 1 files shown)


KittenColony:
Grit has an OD of 10, and is built around making you unable to be slowed down for its duration. Its balanced around having a short metabolism period to prevent macrodosing.

The high OD (20u+) does 20 poison damage for a reason.

No one makes Grit anyway, because Mindleaf is annoying to process, and Bake is significantly more worth-while of an investment.

